PR #7, opened 2021-05-23
added solution

safordog:


src/main/java/mate/controller/driver/AddDriverController.java
```diff
@@ -23,10 +23,10 @@ protected void doGet(HttpServletRequest req, HttpServletResponse resp)
...
             throws IOException, ServletException {
-        String name = req.getParameter("name");
-        String licenceNumber = req.getParameter("licence_number");
-        Driver driver = new Driver(name, licenceNumber);
+        Driver driver = new Driver(req.getParameter("name"),
+                req.getParameter("licence_number"), req.getParameter("login"),
+                req.getParameter("password"));
         driverService.create(driver);
-        resp.sendRedirect("/drivers/add");
+        req.getRequestDispatcher("/login").forward(req, resp);
```
mankokolya, line 30: better to use resp.sendRedirect because after forwarding the parameters you send are saved in the browser and if refreshing the page you will send them again to server

src/main/java/mate/web/filter/AuthenticationFilter.java
```diff
@@ -0,0 +1,45 @@
...
+
+    @Override
+    public void doFilter(ServletRequest servletRequest,
+                         ServletResponse servletResponse, FilterChain filterChain)
+            throws IOException, ServletException {
+        HttpServletRequest req = (HttpServletRequest) servletRequest;
+        HttpServletResponse resp = (HttpServletResponse) servletResponse;
+        HttpSession session = req.getSession();
+        Long driverId = (Long) session.getAttribute("driver_id");
+        if (driverId == null && allowedUrls.contains(req.getServletPath())) {
```
mankokolya, line 35: it will work without driverId == null. You just check if the any user comes from allowed urls you call another filter

mankokolya (changes requested):

src/main/java/mate/web/filter/AuthenticationFilter.java
safordog: Thanks a lot!!!

mankokolya (approved):

Resci (changes requested):

Resci (approved):

PR #10, opened 2021-05-24
first solution

dvdovynets:


src/main/java/mate/controller/driver/GetMyCurrentCarsController.java
```diff
@@ -0,0 +1,26 @@
...
+
+public class GetMyCurrentCarsController extends HttpServlet {
+    private static final Injector injector = Injector.getInstance("mate");
+    private final CarService carService =
+            (CarService) injector.getInstance(CarService.class);
+
+    @Override
+    protected void doGet(HttpServletRequest req, HttpServletResponse resp)
+            throws ServletException, IOException {
+        Long driverId = (Long) req.getSession().getAttribute("driver_id");
```
iedenysenko, line 21: In common mistakes they as us everywhere to make driver id constant :)

iedenysenko (approved):

src/main/java/mate/controller/driver/GetMyCurrentCarsController.java
dvdovynets: My mistake. Thank you!

src/main/java/mate/web/filter/AuthenticationFilter.java
```diff
@@ -0,0 +1,47 @@
...
+    @Override
+    public void doFilter(ServletRequest servletRequest,
+                         ServletResponse servletResponse, FilterChain filterChain)
+            throws IOException, ServletException {
+        HttpServletRequest httpRequest = (HttpServletRequest) servletRequest;
+        HttpServletResponse httpResponse = (HttpServletResponse) servletResponse;
+        HttpSession session = httpRequest.getSession();
+
+        Long driverId = (Long) session.getAttribute(DRIVER_ID);
+        if (driverId == null && allowedUrls.contains(httpRequest.getServletPath())) {
```
MaliukDaria, line 37: Do we need this check here: driverId == null ?

MaliukDaria (changes requested):

src/main/java/mate/web/filter/AuthenticationFilter.java
dvdovynets: Sorry. My mistake. We don't care at that point if id is null or not because urls are allowed

src/main/java/mate/controller/LogoutController.java
```diff
@@ -0,0 +1,16 @@
...
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+public class LogoutController extends HttpServlet {
+    @Override
+    protected void doGet(HttpServletRequest req, HttpServletResponse resp)
+            throws ServletException, IOException {
+        req.getSession().invalidate();
+        resp.sendRedirect("/webapp/WEB-INF/views/index.jsp");
```
MaliukDaria, line 14:
```suggestion
        resp.sendRedirect("/index");
```

MaliukDaria (changes requested):

src/main/java/mate/controller/LogoutController.java
dvdovynets: Done

src/main/java/mate/model/Driver.java
```diff
@@ -54,4 +78,14 @@ public boolean equals(Object o) {
     public int hashCode() {
         return Objects.hash(id, name, licenseNumber);
     }
+
```
MaliukDaria, line 81: Maybe we should add login into equals and hashcode?

MaliukDaria (changes requested):
one more little comment

src/main/java/mate/model/Driver.java
dvdovynets: Done

MaliukDaria (approved):
Good job!

PR #11, opened 2021-05-24
HW WEB SECURITY

divertitoS:


src/main/java/mate/controller/driver/GetMyCurrentCarsController.java
```diff
@@ -0,0 +1,26 @@
...
+
+public class GetMyCurrentCarsController extends HttpServlet {
+    private static final String DRIVER_ID = "driverId";
+    private static final Injector injector = Injector.getInstance("mate");
+    private final CarService carService = (CarService) injector.getInstance(CarService.class);
+
+    @Override
+    protected void doGet(HttpServletRequest req, HttpServletResponse resp)
+            throws ServletException, IOException {
+        Long currentDriverID = (Long) req.getSession().getAttribute(DRIVER_ID);
```
Vlad-Andr, line 21: DRIVER_ID used only once, I think constant is not appropriate here.

src/main/java/mate/controller/driver/LoginController.java
```diff
@@ -0,0 +1,40 @@
...
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import javax.servlet.http.HttpSession;
+import mate.exception.AuthenticationException;
+import mate.lib.Injector;
+import mate.model.Driver;
+import mate.service.AuthenticationService;
+
+public class LoginController extends HttpServlet {
+    private static final String DRIVER_ID = "driverId";
```
Vlad-Andr, line 15: same, unnecessary constant.

Vlad-Andr (approved):

src/main/java/mate/controller/driver/GetMyCurrentCarsController.java
MaliukDaria: driverId should be constant according to checklist: Make driverId a constant field in all classes where you use it.

src/main/java/mate/dao/DriverDaoImpl.java
```diff
@@ -101,11 +106,31 @@ public boolean delete(Long id) {
         }
     }
 
+    @Override
+    public Optional<Driver> findByLogin(String login) {
+        String query = "SELECT * FROM drivers WHERE login = ?";
```
MaliukDaria, line 111: what about deleted drivers?

src/main/java/mate/service/impl/AuthenticationImpl.java
```diff
@@ -0,0 +1,26 @@
...
+import java.util.Optional;
+import mate.exception.AuthenticationException;
+import mate.lib.Injector;
+import mate.lib.Service;
+import mate.model.Driver;
+import mate.service.AuthenticationService;
+import mate.service.DriverService;
+
+@Service
+public class AuthenticationImpl implements AuthenticationService {
```
MaliukDaria, line 12:
```suggestion
public class AuthenticationServiceImpl implements AuthenticationService {
```

src/main/webapp/web.xml
```diff
@@ -105,4 +105,40 @@
...
+    </servlet-mapping>
+
+    <servlet>
+        <servlet-name>getCarsByDriver</servlet-name>
+        <servlet-class>mate.controller.driver.GetMyCurrentCarsController</servlet-class>
+    </servlet>
+    <servlet-mapping>
+        <servlet-name>getCarsByDriver</servlet-name>
+        <url-pattern>/cars/all</url-pattern>
+    </servlet-mapping>
```
MaliukDaria, line 134:
```suggestion
    <servlet>
        <servlet-name>getMyCars</servlet-name>
        <servlet-class>mate.controller.driver.GetMyCurrentCarsController</servlet-class>
    </servlet>
    <servlet-mapping>
        <servlet-name>getMyCars</servlet-name>
        <url-pattern>/cars/current</url-pattern>
    </servlet-mapping>
```

MaliukDaria: /cars/all looks like you want to get all cars

MaliukDaria (changes requested):

src/main/java/mate/service/impl/AuthenticationImpl.java
divertitoS: ++

src/main/webapp/web.xml
divertitoS: ++

src/main/java/mate/dao/DriverDaoImpl.java
divertitoS: fixed

src/main/java/mate/model/Driver.java
```diff
@@ -54,4 +58,20 @@ public boolean equals(Object o) {
     public int hashCode() {
         return Objects.hash(id, name, licenseNumber);
     }
+
```
MaliukDaria, line 61: Do we need to include login into equals and hashCode?

MaliukDaria (changes requested):

src/main/java/mate/model/Driver.java
```diff
@@ -54,4 +58,20 @@ public boolean equals(Object o) {
...
+        this.password = password;
+    }
+
+    public String getLogin() {
+        return login;
+    }
+
+    public void setLogin(String login) {
+        this.login = login;
+    }
```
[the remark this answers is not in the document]
divertitoS: fixed

src/main/java/mate/model/Driver.java
divertitoS: Fixed. Should I use here sensitive info like password or something?

MaliukDaria: You definitely shouldn't include the password in the toString(), and I don't think that you have to include password to equals and hashCode (but this is not very important in this case).

MaliukDaria (approved):
Good job!

PR #30, opened 2021-05-26
Hw

ritakisluhina:


src/main/java/mate/model/Driver.java
```diff
@@ -6,20 +6,29 @@
...
         this.licenseNumber = licenseNumber;
+        this.login = login;
+        this.password = password;
     }
 
-    public Long getId() {
-        return id;
+    public Driver(String name, String licenseNumber) {
+        this.name = name;
+        this.licenseNumber = licenseNumber;
```
Trident1998, line 21: cheklist

Trident1998 (approved):

src/main/java/mate/controller/car/GetCarController.java
```diff
@@ -0,0 +1,31 @@
...
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import javax.servlet.http.HttpSession;
+import mate.lib.Injector;
+import mate.model.Car;
+import mate.service.CarService;
+
+@WebServlet(urlPatterns = "/cars")
```
MaliukDaria, line 15:
```suggestion
@WebServlet(urlPatterns = "/cars/current")
```

src/main/java/mate/filter/AuthenticationFilter.java
```diff
@@ -0,0 +1,42 @@
...
+import javax.servlet.ServletRequest;
+import javax.servlet.ServletResponse;
+import javax.servlet.annotation.WebFilter;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import javax.servlet.http.HttpSession;
+
+@WebFilter(urlPatterns = "/*")
+public class AuthenticationFilter implements Filter {
+    private static Set<String> URL = new HashSet<>();
```
MaliukDaria, line 19: this field is not constant so we can name it regular way
```suggestion
    private static Set<String> allowedUrls = new HashSet<>();
```

MaliukDaria (changes requested):

MaliukDaria (approved):

PR #39, opened 2021-05-27
add solution

KononenkoKonstantin:


Roman0607 (approved):

src/main/java/mate/web/filter/AuthenticationFilter.java
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    @Override
+    public void doFilter(ServletRequest servletRequest, ServletResponse servletResponse,
+                         FilterChain filterChain) throws IOException, ServletException {
+        HttpServletRequest req = (HttpServletRequest) servletRequest;
+        HttpServletResponse resp = (HttpServletResponse) servletResponse;
+        HttpSession session = req.getSession();
+        Long driverId = (Long) session.getAttribute(DRIVER_ID);
+        if (driverId == null && allowedUrls.contains(req.getServletPath())) {
```
MaliukDaria, line 35: do we need this check here "driverId == null" ?

MaliukDaria (changes requested):

src/main/java/mate/web/filter/AuthenticationFilter.java
KononenkoKonstantin: no

MaliukDaria (approved):

PR #69, opened 2021-07-26
HW-web04

docer95:


src/main/java/mate/dao/DriverDaoImpl.java
```diff
@@ -101,11 +105,31 @@ public boolean delete(Long id) {
...
+            ResultSet resultSet = statement.executeQuery();
+            Driver driver = null;
+            if (resultSet.next()) {
+                driver = getDriver(resultSet);
+                return Optional.of(driver);
+            }
+        } catch (SQLException e) {
+            throw new DataProcessingException("Can't find driver by login " + login, e);
+        }
+        return Optional.empty();
```
Basilius-Third, line 123: there is no need for such complication
```suggestion
            Driver driver = null;
            if (resultSet.next()) {
                driver = getDriver(resultSet);
            }
            return Optional.ofNullable(driver);
        } catch (SQLException e) {
            throw new DataProcessingException("Can't find driver by login " + login, e);
        }
```
//or
```suggestion
            if (resultSet.next()) {
                Driver driver = getDriver(resultSet);
                return Optional.of(driver);
            }
        } catch (SQLException e) {
            throw new DataProcessingException("Can't find driver by login " + login, e);
        }
        return Optional.empty();
```

src/main/java/mate/filter/AuthenticationFilter.java
```diff
@@ -0,0 +1,43 @@
...
+        if (driverId == null && allowedUrls.contains(req.getServletPath())) {
+            chain.doFilter(req, resp);
+            return;
+        }
+        if (driverId == null) {
+            resp.sendRedirect("/login");
+            return;
+        }
+        chain.doFilter(req, resp);
+    }
```
Basilius-Third, line 42: image

Basilius-Third (changes requested):

Basilius-Third (approved):

src/main/java/mate/exception/AuthenticationException.java
```diff
@@ -0,0 +1,7 @@
+package mate.exception;
+
+public class AuthenticationException extends RuntimeException {
```
Sofasmile, line 3: what is the difference between checked and unchecked exceptions?
which type of exception should we use here?

src/main/java/mate/service/AuthenticationServiceImpl.java
```diff
@@ -0,0 +1,23 @@
...
+import mate.exception.AuthenticationException;
+import mate.lib.Injector;
+import mate.lib.Service;
+import mate.model.Driver;
+
+@Service
+public class AuthenticationServiceImpl implements AuthenticationService {
+    private static final Injector injector = Injector.getInstance("mate");
+    private final DriverService driverService =
+            (DriverService) injector.getInstance(DriverService.class);
```
Sofasmile, line 13: let's use Injector annotation here

Sofasmile (changes requested):
great!

src/main/java/mate/exception/AuthenticationException.java
docer95: checked - потрібно опрацьовувати на етапі компіляції
unchecked - помилки які виникають в рантаймі

Sofasmile (approved):

PR #87, opened 2021-10-08
Implements login and logout services

osa240:


src/main/java/mate/controller/LoginController.java
```diff
@@ -0,0 +1,43 @@
...
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import javax.servlet.http.HttpSession;
+import mate.exception.AuthenticationException;
+import mate.lib.Injector;
+import mate.model.Driver;
+import mate.service.AuthenticationService;
+
+@WebServlet(urlPatterns = "/login")
```
AlexRozv, line 15: лучше использовать либо аннотации, либо мапинг и не мешать их вместе

src/main/java/mate/controller/LoginController.java
```diff
@@ -0,0 +1,43 @@
...
+    @Override
+    protected void doPost(HttpServletRequest req, HttpServletResponse resp)
+            throws ServletException, IOException {
+        String login = req.getParameter("login");
+        String password = req.getParameter("password");
+        try {
+            Driver driver = authenticationService.login(login, password);
+            HttpSession session = req.getSession();
+            session.setAttribute("driver_id", driver.getId());
+            session.setAttribute("name", driver.getName());
```
AlexRozv, line 36: Make the name of the attribute that you set to the session a constant field in all classes where you use it.
из чеклиста, везде проверь и поправь

src/main/java/mate/controller/IndexController.java
```diff
@@ -5,11 +5,16 @@
 import javax.servlet.http.HttpServlet;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
+import javax.servlet.http.HttpSession;
 
 public class IndexController extends HttpServlet {
+    private HttpSession session;
```
MaliukDaria, line 11: why do we need this variadle as class field?

src/main/java/mate/controller/IndexController.java
```diff
@@ -5,11 +5,16 @@
...
+
     @Override
     protected void doGet(HttpServletRequest req, HttpServletResponse resp)
             throws ServletException, IOException {
+        session = req.getSession();
+        req.setAttribute("driver",session.getAttribute("name"));
```
MaliukDaria, line 17:
```suggestion
        req.setAttribute("driver", session.getAttribute("name"));
```

I think it's better to make "driver" as constant field

src/main/java/mate/controller/LoginController.java
MaliukDaria: agree

src/main/java/mate/controller/driver/GetMyCurrentCarsController.java
```diff
@@ -0,0 +1,26 @@
...
+    private final CarService carService = (CarService) injector
+            .getInstance(CarService.class);
+
+    @Override
+    protected void doGet(HttpServletRequest req, HttpServletResponse resp)
+            throws ServletException, IOException {
+        Long driverId = (Long) req.getSession().getAttribute("driver_id");
+        List<Car> cars = carService.getAllByDriver(driverId);
+        req.setAttribute("cars", cars);
+        req.getRequestDispatcher("/WEB-INF/views/drivers/cars/cars.jsp").forward(req, resp);
```
MaliukDaria, line 24: We can use existed JSP page for displaying all cars here

src/main/java/mate/model/Driver.java
```diff
@@ -5,6 +5,8 @@
 public class Driver {
```
MaliukDaria, line 5: I think we should define `toString()`. method for `Driver` to display info in exception messages

src/main/java/mate/controller/LoginController.java
```diff
@@ -0,0 +1,46 @@
...
+import mate.model.Driver;
+import mate.service.AuthenticationService;
+
+public class LoginController extends HttpServlet {
+    private static final Injector injector = Injector.getInstance("mate");
+    private static final String DRIVER_ID = "driver_id";
+    private static final String NAME = "name";
+    private static final String ERROR_VARIABLE = "error";
+    private static final String LOGIN = "login";
+    private static final String PASSWORD = "password";
```
MaliukDaria, line 20:
```suggestion
    private static final String DRIVER_ID = "driver_id";
```

I think we need to make as constant only "driver_id", because perhaps in the future we could change this parameter, but it is very unlikely that we will need to change "name", "login" and "password" and the fact that we make them constants will not increase the readability of the code.

src/main/java/mate/model/Driver.java
```diff
@@ -47,11 +65,24 @@ public boolean equals(Object o) {
...
+        return Objects.hash(id, name, login, password, licenseNumber);
+    }
+
+    @Override
+    public String toString() {
+        return "Driver{"
+                + "id=" + id
+                + ", name='" + name + '\''
+                + ", login='" + login + '\''
+                + ", password='" + password + '\''
```
MaliukDaria, line 84: It's not safe to define `password` field in `toString()` method

src/main/java/mate/controller/LoginController.java
osa240: understand! thank you!

src/main/java/mate/model/Driver.java
osa240: exactly)

src/main/java/mate/controller/LoginController.java
osa240: "name" also can be change in future, can I to save it?

MaliukDaria (approved):
👍🏻

PR #91, opened 2021-10-13
solution was added

romuth:


src/main/java/mate/dao/DriverDaoImpl.java
```diff
@@ -73,14 +75,15 @@ public Driver create(Driver driver) {
     @Override
     public Driver update(Driver driver) {
         String query = "UPDATE drivers "
-                + "SET name = ?, license_number = ? "
+                + "SET name = ?, license_number = ?, password = ? "
                 + "WHERE id = ? AND is_deleted = FALSE";
```
Koriastras, line 79: I think we need to update login as well

src/main/webapp/WEB-INF/views/login/login.jsp
```diff
@@ -0,0 +1,42 @@
...
+            <input type="password" name="password" form="authenticate" required>
+        </td>
+        <td>
+            <input type="submit" name="submit" form="authenticate">
+        </td>
+    </tr>
+</table>
+<table border="1" class="table_dark">
+    <tr>
+        <th>Or you car register a new driver</th>
```
Koriastras, line 33:
```suggestion
        <th>Or you can register a new driver</th>
```

src/main/java/mate/controller/driver/AddDriverController.java
```diff
@@ -23,9 +23,11 @@ protected void doGet(HttpServletRequest req, HttpServletResponse resp)
...
             throws IOException {
+        String login = req.getParameter("login");
+        String password = req.getParameter("password");
         String name = req.getParameter("name");
         String licenseNumber = req.getParameter("license_number");
-        Driver driver = new Driver(name, licenseNumber);
+        Driver driver = new Driver(name, licenseNumber, login, password);
         driverService.create(driver);
         resp.sendRedirect("/drivers/add");
     }
```
Koriastras, line 33: if you use @WebServlet annotation, it would be better to use it in each controller

src/main/java/mate/dao/DriverDaoImpl.java
```diff
@@ -105,8 +109,28 @@ private Driver getDriver(ResultSet resultSet) throws SQLException {
...
+        Driver driver = new Driver(name, licenseNumber, login, password);
         driver.setId(id);
         return driver;
     }
+
+    @Override
+    public Optional<Driver> findByLogin(String login) {
+        String findByLoginQuery = "SELECT * FROM drivers WHERE login = ? AND is_deleted = false;";
+        try (Connection connection = ConnectionUtil.getConnection();
+                PreparedStatement preparedStatement =
```
Max-Pochepets, line 123:
```suggestion
                PreparedStatement findDriverByLoginStatement =
```

PR #95, opened 2021-10-14
Authentication made + LoginController

netiazhevskyi:


src/main/java/mate/controller/GetMyCurrentCarsController.java
```diff
@@ -0,0 +1,26 @@
...
+
+public class GetMyCurrentCarsController extends HttpServlet {
+    private static final Injector injector = Injector.getInstance("mate");
+    private final CarService carService = (CarService) injector
+            .getInstance(CarService.class);
+
+    @Override
+    protected void doGet(HttpServletRequest req, HttpServletResponse resp)
+            throws ServletException, IOException {
+        Long sessionId = (Long) req.getSession().getAttribute("user_id");
```
Yehor-Zapara, line 21: Let's make **user_id** constant. And I think it would be better to rename variable **sessionId** to driverId or userId

src/main/java/mate/controller/LoginController.java
```diff
@@ -0,0 +1,40 @@
...
+    @Override
+    protected void doPost(HttpServletRequest req, HttpServletResponse resp)
+            throws ServletException, IOException {
+        String login = req.getParameter("login");
+        String password = req.getParameter("password");
+        try {
+            Driver driver = authenticationService.login(login, password);
+            resp.sendRedirect("/index");
+            HttpSession session = req.getSession();
+            session.setAttribute("user_id", driver.getId());
```
Yehor-Zapara, line 34: The same about constant. It was in the checklist

src/main/java/mate/dao/DriverDaoImpl.java
```diff
@@ -101,11 +103,30 @@ public boolean delete(Long id) {
         }
     }
 
+    @Override
+    public Optional<Driver> findByLogin(String login) {
+        String query = "SELECT * FROM drivers WHERE login = ?";
```
Yehor-Zapara, line 108: Let's check for **is_deleted** is FALSE

src/main/java/mate/dao/DriverDaoImpl.java
```diff
@@ -17,13 +17,15 @@
 public class DriverDaoImpl implements DriverDao {
```
Yehor-Zapara, line 17: Let's add columns login and password in **update** method

src/main/java/mate/web/AuthenticationFilter.java
```diff
@@ -0,0 +1,42 @@
...
+        allowedUrls.add("/drivers/add");
+    }
+
+    @Override
+    public void doFilter(ServletRequest request, ServletResponse response, FilterChain chain)
+            throws IOException, ServletException {
+        HttpServletRequest req = (HttpServletRequest) request;
+        HttpServletResponse resp = (HttpServletResponse) response;
+        HttpSession session = req.getSession();
+        Long userId = (Long) session.getAttribute("user_id");
```
Yehor-Zapara, line 31: Same about constant

src/main/java/mate/web/AuthenticationFilter.java
```diff
@@ -0,0 +1,42 @@
...
+        if (userId == null && allowedUrls.contains(req.getServletPath())) {
+            chain.doFilter(req, resp);
+            return;
+        }
+        if (userId == null) {
+            resp.sendRedirect("/login");
+            return;
+        }
```
Yehor-Zapara, line 39: Let's use one **if** statement. We can use userId == null && !allowedUrls.contains(req.getServletPath()) in condition

src/main/webapp/web.xml
```diff
@@ -16,6 +32,15 @@
         <url-pattern>/</url-pattern>
     </servlet-mapping>
 
+    <servlet>
+        <servlet-name>getMyCurrentCarsController</servlet-name>
+        <servlet-class>mate.controller.GetMyCurrentCarsController</servlet-class>
+    </servlet>
+    <servlet-mapping>
+        <servlet-name>getMyCurrentCarsController</servlet-name>
+        <url-pattern>/cars/myCars</url-pattern>
```
Yehor-Zapara, line 41: Maybe it is better to use **/cars/my**? But it's only advice

src/main/java/mate/model/Driver.java
```diff
@@ -6,12 +6,36 @@
     private Long id;
     private String name;
     private String licenseNumber;
+    private String login;
+    private String password;
 
     public Driver(String name, String licenseNumber) {
```
MaliukDaria, line 12: I think we should add toString() method here(to display valid data in exception messages)

src/main/webapp/web.xml
MaliukDaria, line 41: agree, lets change mapping url to /cars/my or cars/current

PR #97, opened 2021-10-14
added AuthenticationFilter to web project. Implemented LoginController, LogoutController and ReturnBackController to return to index page.

MiaHrynda:


18Yuliya18 (approved):

src/main/java/mate/controller/ReturnBackController.java
```diff
@@ -0,0 +1,16 @@
+package mate.controller;
+
+import java.io.IOException;
+import javax.servlet.ServletException;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+public class ReturnBackController extends HttpServlet {
```
MaliukDaria, line 9: How the functionality of this controller differs from Index controller?

src/main/java/mate/model/Driver.java
```diff
@@ -5,13 +5,21 @@
 public class Driver {
     private Long id;
     private String name;
+    private String login;
+    private String password;
     private String licenseNumber;
 
     public Driver(String name, String licenseNumber) {
```
MaliukDaria, line 12: I think we should add `toString()` method here(to display valid data in exception messages)

MaliukDaria (changes requested):

PR #99, opened 2021-10-14
web-security

dogovor:


src/main/webapp/WEB-INF/views/drivers/add.jsp
```diff
@@ -13,13 +13,21 @@
...
     <tr>
         <td>
             <input type="text" name="name" form="driver" required>
         </td>
+        <td>
+            <input type="text" name="login" form="driver" required>
+        </td>
+        <td>
+            <input type="text" name="password" form="driver" required>
+        </td>
```
Zbarazkyy, line 30: I would add fields Login and Password so everyone knows your login and password

src/main/java/mate/controller/driver/AddDriverController.java
```diff
@@ -24,8 +24,10 @@ protected void doGet(HttpServletRequest req, HttpServletResponse resp)
     protected void doPost(HttpServletRequest req, HttpServletResponse resp)
             throws IOException {
         String name = req.getParameter("name");
+        String login = req.getParameter("name");
+        String password = req.getParameter("name");
         String licenseNumber = req.getParameter("license_number");
```
Zbarazkyy, line 29: I would add fields Login and Password so everyone knows your login and password

Zbarazkyy (changes requested):
The problem with Tevis

PR #102, opened 2021-10-16
changed Driver class, pom.xml, web.xml, index.jsp, add.jsp. Created authentication filter.

sich-mykhailo:


MAFIOZNIK1999 (changes requested):
а где logout.jsp?

src/main/java/mate/controller/LogoutController.java
```diff
@@ -0,0 +1,16 @@
...
+import javax.servlet.http.HttpServletResponse;
+
+public class LogoutController extends HttpServlet {
+    @Override
+    protected void doGet(HttpServletRequest req, HttpServletResponse resp)
+            throws ServletException, IOException {
+        req.getSession().invalidate();
+        resp.sendRedirect("/login");
+    }
+}
```
MAFIOZNIK1999, line 16: added logout.jsp file

MAFIOZNIK1999 (changes requested):

src/main/java/mate/controller/LogoutController.java
sich-mykhailo: i created a "logout" button in the index.jsp file, I think i don't need another file.

MAFIOZNIK1999: пусть лучше ментор тогда скажет

MAFIOZNIK1999 (approved):

PR #106, opened 2021-10-17
DmytroPetrushenko / jv-web-security Public

DmytroPetrushenko:


.circleci/config.yml
```diff
@@ -0,0 +1,36 @@
+# Use the latest 2.1 version of CircleCI pipeline process engine.
```
RomanDubovskyi, line 1: do we need this file?

src/main/java/mate/service/AuthenticationServiceImpl.java
```diff
@@ -0,0 +1,25 @@
...
+    @Inject
+    private DriverService driverService;
+
+    @Override
+    public Driver login(String login, String password) throws AuthenticationException {
+        Optional<Driver> driver = driverService.findByLogin(login);
+        if (driver.isEmpty()) {
+            throw new AuthenticationException("Login and password were incorrect!");
+        }
+        if (driver.get().getPassword().equals(password)) {
```
RomanDubovskyi, line 20: we can modify this if() so we won't need second `if (driver.isEmpty()) {`

src/main/java/mate/web/filter/AuthenticationFilter.java
```diff
@@ -0,0 +1,45 @@
...
+
+    @Override
+    public void doFilter(ServletRequest request, ServletResponse response, FilterChain chain)
+            throws IOException, ServletException {
+        HttpServletRequest req = (HttpServletRequest) request;
+        HttpServletResponse resp = (HttpServletResponse) response;
+        HttpSession session = req.getSession();
+
+        Long driveId = (Long) session.getAttribute("driverId");
+        if (driveId == null && allowedUrl.contains(req.getServletPath())) {
```
RomanDubovskyi, line 33: same here, one if block with correct condition is enough

PR #110, opened 2021-10-17
jv-web-security task completed

Gleb-Kreshchuk:


src/main/webapp/WEB-INF/views/cars/add.jsp
```diff
@@ -8,6 +8,7 @@
     <title>Add car</title>
 </head>
 <body>
+<jsp:include page="/WEB-INF/views/header/header.jsp"/>
```
KyryloDvornyk, line 11: Why do we need register or login when we have already logined?

src/main/java/mate/controller/LoginController.java
```diff
@@ -0,0 +1,43 @@
...
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import javax.servlet.http.HttpSession;
+import mate.exception.AuthenticationException;
+import mate.lib.Injector;
+import mate.model.Driver;
+import mate.service.AuthenticationService;
+
+@WebServlet(urlPatterns = "/login")
```
KyryloDvornyk, line 15: I think its better to write it all in web.xml or all with annotations

KyryloDvornyk:
Add your GetMyCurrentCarsController servlet call in index.jsp

src/main/webapp/WEB-INF/views/cars/add.jsp
Gleb-Kreshchuk: for the sake of logout

MAFIOZNIK1999 (approved):

src/main/java/mate/controller/LoginController.java
Resci: Я за те, щоб для фільтра також зробити конфігурування анотацією

Gleb-Kreshchuk: Done

PR #133, opened 2021-11-12
added solution

YaroslavYarmoshyk:


src/main/java/mate/controller/driver/GetMyCurrentCarsController.java
```diff
@@ -0,0 +1,15 @@
+package mate.controller.driver;
+
+import java.io.IOException;
+import javax.servlet.ServletException;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+public class GetMyCurrentCarsController extends HttpServlet {
```
RomanDubovskyi, line 9: do we use it anywhere?

src/main/java/mate/service/AuthenticationServiceImpl.java
```diff
@@ -0,0 +1,31 @@
...
+public class AuthenticationServiceImpl implements AuthenticationService {
+    @Inject
+    private DriverService driverService;
+
+    @Override
+    public Driver login(String username, String password) throws AuthenticationException {
+        Optional<Driver> driver;
+        try {
+            driver = driverService.findByLogin(username);
+        } catch (DataProcessingException e) {
```
RomanDubovskyi, line 20: let's use single if construction in this method, plus you don't need to have try catch. If code is working fine you won't have DataProcessingException. But if it happens it means SQL code is wrong and we need to fix it

src/main/java/mate/web/filter/AuthFilter.java
```diff
@@ -0,0 +1,46 @@
...
+
+    @Override
+    public void doFilter(ServletRequest servletRequest, ServletResponse servletResponse,
+                         FilterChain filterChain) throws IOException, ServletException {
+        HttpServletRequest request = (HttpServletRequest) servletRequest;
+        HttpServletResponse response = (HttpServletResponse) servletResponse;
+        HttpSession session = request.getSession();
+        Long driverId = (Long) session.getAttribute("driver_id");
+
+        if (driverId == null && allowedUrl.contains(request.getServletPath())) {
```
RomanDubovskyi, line 29: le'ts use single if clause here as well to shorten the code

RomanDubovskyi (changes requested):

RomanDubovskyi (approved):

PR #134, opened 2021-12-02
added two fields in to driver model;

forsfakel:
correct jsp page create driver for new fields;
correct DriverDaoImpl and CarDaoImpl;
created new page and controller for enter to service;
created AuthenticationService;
created AuthenticationFilter for some drivers can not see to other services;

### DIFF
--- a/src/main/java/mate/service/AuthenticationServiceImpl.java
+++ b/src/main/java/mate/service/AuthenticationServiceImpl.java
@@ -1,0 +1,25 @@
+package mate.service;
+
+import java.util.Optional;
+import mate.dao.DriverDao;
+import mate.exception.AuthenticationException;
+import mate.lib.Inject;
+import mate.lib.Service;
+import mate.model.Driver;
+
+@Service
+public class AuthenticationServiceImpl implements AuthenticationService {
+    @Inject private DriverDao driverDao;
+
+    @Override
+    public Driver login(String login, String password) throws AuthenticationException {
+        Optional<Driver> driver = driverDao.findByLogin(login);
+        if (driver.isEmpty()) {
+            throw new AuthenticationException("Login or password was incorrect");
+        }
+        if (driver.get().getPassword().equals(password)) {
+            return driver.get();
+        }
+        throw new AuthenticationException("Login or password was incorrect");
+    }
+}
--- a/src/main/java/mate/web/filter/AuthenticationFilter.java
+++ b/src/main/java/mate/web/filter/AuthenticationFilter.java
@@ -1,0 +1,32 @@
+package mate.web.filter;
+
+import java.io.IOException;
+import javax.servlet.Filter;
+import javax.servlet.FilterChain;
+import javax.servlet.ServletException;
+import javax.servlet.ServletRequest;
+import javax.servlet.ServletResponse;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import javax.servlet.http.HttpSession;
+
+public class AuthenticationFilter implements Filter {
+    @Override
+    public void doFilter(ServletRequest servletRequest, ServletResponse servletResponse,
+            FilterChain filterChain) throws IOException, ServletException {
+        HttpServletRequest req = (HttpServletRequest) servletRequest;
+        HttpServletResponse resp = (HttpServletResponse) servletResponse;
+        HttpSession session = req.getSession();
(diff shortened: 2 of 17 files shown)


src/main/java/mate/service/AuthenticationServiceImpl.java
```diff
@@ -0,0 +1,25 @@
...
+import java.util.Optional;
+import mate.dao.DriverDao;
+import mate.exception.AuthenticationException;
+import mate.lib.Inject;
+import mate.lib.Service;
+import mate.model.Driver;
+
+@Service
+public class AuthenticationServiceImpl implements AuthenticationService {
+    @Inject private DriverDao driverDao;
```
RomanDubovskyi, line 12:
```suggestion
    @Inject 
    private DriverDao driverDao;
```

src/main/java/mate/service/AuthenticationServiceImpl.java
```diff
@@ -0,0 +1,25 @@
...
+
+    @Override
+    public Driver login(String login, String password) throws AuthenticationException {
+        Optional<Driver> driver = driverDao.findByLogin(login);
+        if (driver.isEmpty()) {
```
RomanDubovskyi, line 17: single if statement is enough here if you modify check inside of it

src/main/java/mate/web/filter/AuthenticationFilter.java
```diff
@@ -0,0 +1,32 @@
...
+public class AuthenticationFilter implements Filter {
+    @Override
+    public void doFilter(ServletRequest servletRequest, ServletResponse servletResponse,
+            FilterChain filterChain) throws IOException, ServletException {
+        HttpServletRequest req = (HttpServletRequest) servletRequest;
+        HttpServletResponse resp = (HttpServletResponse) servletResponse;
+        HttpSession session = req.getSession();
+        Long driverId = (Long) session.getAttribute("driver_id");
+        if (driverId == null && req.getServletPath().equals("/login") || req.getServletPath()
+                .equals("/drivers/add")) {
```
RomanDubovskyi, line 22: let's add these endpoints into set, in case there will be more of them in the future 
\+ we can use single if here as well

RomanDubovskyi (changes requested):

src/main/java/mate/service/AuthenticationServiceImpl.java
forsfakel: +

src/main/java/mate/web/filter/AuthenticationFilter.java
forsfakel: +

src/main/java/mate/web/filter/AuthenticationFilter.java
```diff
@@ -0,0 +1,33 @@
...
+    @Override
+    public void doFilter(ServletRequest servletRequest, ServletResponse servletResponse,
+            FilterChain filterChain) throws IOException, ServletException {
+        HttpServletRequest req = (HttpServletRequest) servletRequest;
+        HttpServletResponse resp = (HttpServletResponse) servletResponse;
+        HttpSession session = req.getSession();
+        Long driverId = (Long) session.getAttribute("driver_id");
+        Set<String> links = new HashSet<>();
+        links.add("/login");
+        links.add("/drivers/add");
```
MaliukDaria, line 26: Lets move this Set to class level

src/main/webapp/WEB-INF/views/drivers/add.jsp
```diff
@@ -23,6 +25,12 @@
         <td>
             <input type="text" name="license_number" form="driver" required>
         </td>
+        <td>
+            <input type="text" name="login" form="driver" required>
+        </td>
+        <td>
+            <input type="text" name="password" form="driver" required>
```
MaliukDaria, line 32:
```suggestion
            <input type="password" name="password" form="driver" required>
```

MaliukDaria (changes requested):

src/main/java/mate/web/filter/AuthenticationFilter.java
forsfakel: + use method init()

src/main/webapp/WEB-INF/views/drivers/add.jsp
forsfakel: +

MaliukDaria (approved):
Good Job 👍🏻

PR #154, opened 2021-12-14
jv-web-security solution

Savujkee:


PineappleApplePinecone (changes requested):

src/main/java/mate/service/AuthenticationServiceImpl.java
```diff
@@ -0,0 +1,25 @@
...
+
+@Service
+public class AuthenticationServiceImpl implements AuthenticationService {
+    @Inject
+    private DriverDao driverDao;
+
+    public Driver login(String login, String password) throws AuthenticationException {
+        Optional<Driver> driver = driverDao.findByLogin(login);
+        if (driver.isEmpty()) {
+            throw new AuthenticationException("Login or password is invalid");
```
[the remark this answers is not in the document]
Savujkee: Nope, we don't want to specify which particular parameter is invalid.
I use 2 consequent if blocks here instead of a single if with &&.
it doesn't change algorithm complexity but adds a few unnecessary lines.
Gonna change it)

PineappleApplePinecone (approved):

src/main/java/mate/web/filter/AuthenticationFilter.java
```diff
@@ -0,0 +1,44 @@
...
+        Long driverId = (Long) session.getAttribute("id");
+        if (driverId == null && allowedUrls.contains(req.getServletPath())) {
+            filterChain.doFilter(req, resp);
+            return;
+        }
+        if (driverId == null) {
+            resp.sendRedirect("/login");
+            return;
+        }
+        filterChain.doFilter(req, resp);
```
AlexRozv, line 42: can we simplify it?
Screenshot 2021-12-14 at 09 45 55

AlexRozv (changes requested):

src/main/java/mate/web/filter/AuthenticationFilter.java
Savujkee: yes

AlexRozv (approved):
good job 👍

PR #179, opened 2021-12-22
make creation of objects be consistent

issugy:
chose one option for creation of objects in methods for conversion from ResultSet - setters

### DIFF
--- a/src/main/java/taxi/dao/CarDaoImpl.java
+++ b/src/main/java/taxi/dao/CarDaoImpl.java
@@ -226,8 +226,10 @@ public class CarDaoImpl implements CarDao {
         Long driverId = resultSet.getObject("id", Long.class);
         String name = resultSet.getNString("name");
         String licenseNumber = resultSet.getNString("license_number");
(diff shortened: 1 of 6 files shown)


src/main/java/mate/model/Car.java
```diff
@@ -10,6 +10,9 @@
     private Manufacturer manufacturer;
     private List<Driver> drivers;
 
+    public Car() {
+    }
+
     public Car(String model, Manufacturer manufacturer) {
```
Nick97-git, line 16: Is this constructor used anywhere? the same question about other models

Nick97-git:
Also, I don't see PR for fixing this problem in `jv-web-practice`. You didn't notice that it should be fixed too or it's in progress?

src/main/java/mate/model/Car.java
issugy: Снимок экрана 2021-12-23 в 10 26 17
it's used in the doPost() method, same for every other models

Nick97-git (approved):

PR #180, opened 2021-12-22
make methods' names be consistent

issugy:
make names of methods for conversion from ResultSet in dao implementations be consistent


Nick97-git (approved):

Nick97-git:
Also, I have the same question as in the previous PR

PR #181, opened 2022-01-10
Wrote an assignment jv-web-security.

ASelivonchyk:
- added fields "login" and "password" to Driver model;
- modified AddDriverController with field "login" and "password";
- modified drivers/add.jsp to have ability adding "login" and "password";
- modified drivers/all.jsp and cars/all.jsp to show driver "login";
- modified DriverDaoImpl and CarDaoImpl to fetch all drivers columns from DB;
- in DriverDao and DriverService added new method findByLogin();
- created AuthenticationService and it Impl to login driver in taxi service;
- made LoginController with corresponding login.jsp page;
- made LogoutController with header.jsp and added button "logout" to index.jsp ;
- implemented AuthenticationFilter that provide ability to visit only /login and /drivers/add pages to not authorized users;
- created AuthenticationException;
- implemented GetMyCurrentCarsController with ability to registered drivers see its cars;
- moved all implementation to Impl packages;
- replaced mapping in web.xml with web annotations;


src/main/java/taxi/web/filter/AuthenticationFilter.java
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    @Override
+    public void doFilter(ServletRequest request, ServletResponse response, FilterChain chain)
+            throws IOException, ServletException {
+        HttpServletRequest req = (HttpServletRequest) request;
+        HttpServletResponse resp = (HttpServletResponse) response;
+        HttpSession session = req.getSession();
+        Long driverId = (Long) session.getAttribute("driver_id");
+        if (driverId == null && allowedUrls.contains(req.getServletPath())) {
```
RomanDubovskyi, line 34: we can use single if here

RomanDubovskyi (approved):

PR #183, opened 2022-02-02
added AuthenticationFilter and realized LoginController, LogoutContro…

ev-lanovenko:
…ller

### DIFF
--- a/src/main/java/taxi/web/filter/AuthenticationFilter.java
+++ b/src/main/java/taxi/web/filter/AuthenticationFilter.java
@@ -1,0 +1,38 @@
+package taxi.web.filter;
+
+import java.io.IOException;
+import java.util.HashSet;
+import java.util.Set;
+import javax.servlet.*;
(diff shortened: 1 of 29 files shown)


src/main/java/taxi/web/filter/AuthenticationFilter.java
```diff
@@ -0,0 +1,43 @@
...
+        Long driverId = (Long) session.getAttribute("driver_id");
+        if (driverId == null && allowedUrls.contains(req.getServletPath())) {
+            filterChain.doFilter(req, resp);
+            return;
+        }
+        if (driverId == null) {
+            resp.sendRedirect("/login");
+            return;
+        }
+        filterChain.doFilter(req, resp);
```
MaliukDaria, line 41: can we do this only with one if?

MaliukDaria (changes requested):

MaliukDaria (approved):

PR #189, opened 2022-02-21
Hw web security main

maxkmaxk:


src/main/java/taxi/service/AuthenticationServiceImpl.java
```diff
@@ -0,0 +1,25 @@
...
+
+    @Override
+    public Driver login(String login, String password) throws AuthenticationException {
+        Optional<Driver> driver = driverService.findByLogin(login);
+        if (driver.isEmpty()) {
+            throw new AuthenticationException("Username or password was incorrect");
+        }
+        if (driver.get().getPassword().equals(password)) {
+            return driver.get();
+        }
```
MaliukDaria, line 22: can we use only one if statement here?

MaliukDaria (changes requested):

MaliukDaria (approved):

PR #192, opened 2022-02-23
web security (login log out) service realisation

LinkClink:


src/main/java/taxi/controller/LogOutController.java
```diff
@@ -0,0 +1,16 @@
+package taxi.controller;
+
+import java.io.IOException;
+import javax.servlet.ServletException;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+public class LogOutController extends HttpServlet {
```
MaliukDaria, line 9:
```suggestion
public class LogoutController extends HttpServlet {
```

src/main/java/taxi/controller/LogOutController.java
```diff
@@ -0,0 +1,16 @@
...
+    @Override
+    protected void doGet(HttpServletRequest req, HttpServletResponse resp)
+            throws ServletException, IOException {
+        req.getSession().invalidate();
+        resp.sendRedirect("/index");
```
MaliukDaria, line 14: If you use sendRedirect() method in your controllers, please pass "request.getContextPath()" + "/your-endpoint" as a parameter. Currently, the context path is empty, but if it is not, your code still should work.

src/main/java/taxi/service/AuthenticationServiceImpl.java
```diff
@@ -0,0 +1,24 @@
...
+
+    @Override
+    public Driver login(String login, String password) throws AuthenticationException {
+        Driver exitDriver = driverService.findByLogin(login);
+        if (exitDriver == null) {
+            throw new AuthenticationException("Wrong password or login");
+        } else if (exitDriver.getPassword().equals(password)) {
+            return exitDriver;
+        }
+        throw new AuthenticationException("Wrong password or login");
```
MaliukDaria, line 22: could we use only one if here?

src/main/java/taxi/web/filter/AuthenticationFilter.java
```diff
@@ -0,0 +1,34 @@
...
+        Object userId = session.getAttribute("user_id");
+        if (userId == null && req.getServletPath().equals("/login")) {
+            chain.doFilter(req, res);
+            return;
+        }
+        if (userId == null) {
+            res.sendRedirect("/login");
+            return;
+        }
+        chain.doFilter(req, res);
```
MaliukDaria, line 32: could we use only one if here?

src/main/java/taxi/web/filter/AuthenticationFilter.java
```diff
@@ -0,0 +1,34 @@
...
+public class AuthenticationFilter implements Filter {
+
+    @Override
+    public void doFilter(ServletRequest request, ServletResponse response, FilterChain chain)
+            throws ServletException, IOException {
+        HttpServletRequest req = (HttpServletRequest) request;
+        HttpServletResponse res = (HttpServletResponse) response;
+        HttpSession session = req.getSession();
+        Object userId = session.getAttribute("user_id");
+        if (userId == null && req.getServletPath().equals("/login")) {
```
MaliukDaria, line 24: lets create some Set of allowed urls(we can add more urls to this Set in the future without changing other code)

MaliukDaria (changes requested):

MaliukDaria (approved):
👍🏻

PR #196, opened 2022-03-12
Driver login system was implemented

karma-o:


src/main/java/taxi/controller/driver/LoginController.java
```diff
@@ -0,0 +1,42 @@
...
+    protected void doPost(HttpServletRequest req, HttpServletResponse resp)
+            throws IOException {
+        String login = req.getParameter("login");
+        String password = req.getParameter("password");
+        try {
+            Driver driver = authService.login(login, password);
+            req.getSession().setAttribute("driver_id", driver.getId());
+            resp.sendRedirect(req.getContextPath() + "/index");
+        } catch (AuthenticationException e) {
+            throw new RuntimeException("invalid login or password, try again");
```
KostyaNikiforov, line 39:
```suggestion
            throw new RuntimeException(e.getMessage());
```
If we are already throwing an exception with a custom message, we can reuse it this way.

KostyaNikiforov (approved):

PR #205, opened 2022-03-25
authentication impl

anyutamysko:


src/main/java/taxi/model/Driver.java
```diff
@@ -39,6 +52,22 @@ public void setLicenseNumber(String licenseNumber) {
...
+
+    public String getPassword() {
+        return password;
+    }
+
+    public void setPassword(String password) {
+        this.password = password;
+    }
+
     @Override
```
sltnsrh, line 71: If we have methods equals and hashCode, we need to add our new fields to them.

sltnsrh (changes requested):

src/main/java/taxi/model/Driver.java
```diff
@@ -79,11 +79,13 @@ public boolean equals(Object o) {
         Driver driver = (Driver) o;
         return Objects.equals(id, driver.id)
                 && Objects.equals(name, driver.name)
-                && Objects.equals(licenseNumber, driver.licenseNumber);
+                && Objects.equals(licenseNumber, driver.licenseNumber)
+                && Objects.equals(login, driver.login)
+                && Objects.equals(password, password);
```
sltnsrh, line 84: && Objects.equals(password, driver.password);

sltnsrh (changes requested):

sltnsrh (approved):

KostyaNikiforov (approved):

PR #206, opened 2022-03-27
web-security implementation

sltnsrh:


src/main/java/taxi/controller/driver/LoginDriverController.java
```diff
@@ -0,0 +1,40 @@
...
+    @Override
+    protected void doPost(HttpServletRequest req, HttpServletResponse resp)
+            throws ServletException, IOException {
+        String login = req.getParameter("login");
+        String password = req.getParameter("password");
+        try {
+            Driver driver = authenticationService.login(login, password);
+            HttpSession session = req.getSession();
+            session.setAttribute("driver_id", driver.getId());
+            resp.sendRedirect("/index");
```
anyutamysko, line 34: resp.sendRedirect(req.getContextPath() + "/index");

src/main/java/taxi/controller/driver/AddDriverController.java
```diff
@@ -25,8 +26,15 @@ protected void doPost(HttpServletRequest req, HttpServletResponse resp)
             throws IOException {
         String name = req.getParameter("name");
         String licenseNumber = req.getParameter("license_number");
-        Driver driver = new Driver(name, licenseNumber);
+        String login = req.getParameter("login");
+        String password = req.getParameter("password");
+        Driver driver = new Driver(name, licenseNumber, login, password);
         driverService.create(driver);
+        HttpSession session = req.getSession();
+        if (session.getAttribute("driver_id") == null) {
```
anyutamysko, line 34: I think, it would be better to  check in filter

sltnsrh: For example?
Just used session driver id for redirect logic after a new driver registration.

anyutamysko (approved):

PR #221, opened 2022-04-04
Implement authentication in taxi service

ValeraLemberov:


src/main/java/taxi/controller/LogoutController.java
```diff
@@ -0,0 +1,18 @@
...
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+@WebServlet(urlPatterns = "/logout")
+public class LogoutController extends HttpServlet {
+    @Override
+    protected void doGet(HttpServletRequest req, HttpServletResponse resp)
+            throws ServletException, IOException {
+        req.getSession().invalidate();
+        req.getRequestDispatcher("WEB-INF/views/index.jsp").forward(req,resp);
```
MaliukDaria, line 16: What is the difference between sendRedirect and forward?

src/main/java/taxi/dao/DriverDaoImpl.java
```diff
@@ -111,4 +115,24 @@ private Driver parseDriverFromResultSet(ResultSet resultSet) throws SQLException
         driver.setLicenseNumber(licenseNumber);
         return driver;
     }
+
+    @Override
+    public Optional<Driver> findByLogin(String login) {
```
MaliukDaria, line 120: public methods should be above private

src/main/java/taxi/dao/DriverDaoImpl.java
```diff
@@ -111,4 +115,24 @@ private Driver parseDriverFromResultSet(ResultSet resultSet) throws SQLException
...
+                + " WHERE login = ? AND is_deleted = FALSE;";
+        try (Connection connection = ConnectionUtil.getConnection();
+                PreparedStatement preparedStatement = connection.prepareStatement(query)) {
+            preparedStatement.setString(1,login);
+            ResultSet resultSet = preparedStatement.executeQuery();
+            Driver driver = null;
+            if (resultSet.next()) {
+                driver = parseDriverFromResultSet(resultSet);
+                driver.setLogin(resultSet.getString("login"));
+                driver.setPassword(resultSet.getString("password"));
```
MaliukDaria, line 131: lets add this logic into parseDriverFromResultSet() method (I think we should get all info about driver in all methods)

src/main/java/taxi/model/Driver.java
```diff
@@ -6,13 +6,38 @@
     private Long id;
     private String name;
     private String licenseNumber;
+    private String login;
+    private String password;
 
     public Driver() {
```
MaliukDaria, line 12: Lets define toString() method to display driver instances properly in exception messages(be attentive to sensitive driver fields)

src/main/java/taxi/service/AuthenticationServiceImpl.java
```diff
@@ -0,0 +1,31 @@
...
+import taxi.dao.DriverDaoImpl;
+import taxi.exception.AuthenticationException;
+import taxi.lib.Inject;
+import taxi.lib.Service;
+import taxi.model.Driver;
+
+@Service
+public class AuthenticationServiceImpl implements AuthenticationService {
+    @Inject
+    private DriverDao driverDao = new DriverDaoImpl();
```
MaliukDaria, line 14: Why do we use @Inject annotation?

src/main/java/taxi/service/DriverServiceImpl.java
```diff
@@ -2,14 +2,15 @@
...
+import taxi.dao.DriverDaoImpl;
 import taxi.lib.Inject;
 import taxi.lib.Service;
 import taxi.model.Driver;
 
 @Service
 public class DriverServiceImpl implements DriverService {
     @Inject
-    private DriverDao driverDao;
+    private DriverDao driverDao = new DriverDaoImpl();
```
MaliukDaria, line 13: same, why do we use  @Inject annotation?

MaliukDaria (changes requested):

PR #225, opened 2022-04-19
Solution of the task.

k1llzers:


src/main/java/taxi/filter/AuthenticationFilter.java
```diff
@@ -0,0 +1,45 @@
...
+        HttpSession session = req.getSession();
+        Long driverId = (Long) session.getAttribute("driver_id");
+        if (driverId == null && allowedUrls.contains(req.getServletPath())) {
+            filterChain.doFilter(req,resp);
+            return;
+        }
+        if (driverId == null) {
+            resp.sendRedirect(req.getContextPath() + "/login");
+            return;
+        }
```
MaliukDaria, line 42: can we do this only with one if

src/main/java/taxi/service/AuthenticationServiceImpl.java
```diff
@@ -0,0 +1,23 @@
...
+public class AuthenticationServiceImpl implements AuthenticationService {
+    @Inject
+    private DriverService driverService;
+
+    @Override
+    public Driver login(String login, String password) throws AuthenticationException {
+        Optional<Driver> driver = driverService.findByLogin(login);
+        if (driver.isEmpty() || !driver.get().getPassword().equals(password)) {
+            throw new AuthenticationException("Username or password was incorrect");
+        } else {
```
MaliukDaria, line 19: do we need else after throw?

MaliukDaria (changes requested):

MaliukDaria (approved):

PR #238, opened 2022-05-18
task web-security

MarkShapka:
- created AuthenticationService()
- added header.jsp, login.jsp pages
- implemented filter based on HttpSession

### DIFF
--- a/src/main/java/taxi/service/AuthenticationServiceImpl.java
+++ b/src/main/java/taxi/service/AuthenticationServiceImpl.java
@@ -1,0 +1,25 @@
+package taxi.service;
+
+import java.util.Optional;
+import taxi.exception.AuthenticationException;
+import taxi.lib.Inject;
+import taxi.lib.Service;
+import taxi.model.Driver;
+
+@Service
+public class AuthenticationServiceImpl implements AuthenticationService {
+    @Inject
+    private DriverService driverService;
+
+    @Override
+    public Driver login(String login, String password) throws AuthenticationException {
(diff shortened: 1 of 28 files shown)


src/main/java/taxi/service/AuthenticationServiceImpl.java
```diff
@@ -0,0 +1,25 @@
...
+
+    @Override
+    public Driver login(String login, String password) throws AuthenticationException {
+        Optional<Driver> driverFromDB = driverService.findByLogin(login);
+        if (driverFromDB.isEmpty()) {
+            throw new AuthenticationException("Login or password are incorrect");
+        }
+        if (driverFromDB.get().getPassword().equals(password)) {
+            return driverFromDB.get();
+        }
```
MaliukDaria, line 22: could we use only one if statement here?

src/main/java/taxi/service/filter/AuthenticationFilter.java
```diff
@@ -0,0 +1,42 @@
...
+        HttpSession session = req.getSession();
+        String driverLogin = (String) session.getAttribute("driver_login");
+        if (driverLogin == null && allowedUrl.contains(req.getServletPath())) {
+            filterChain.doFilter(req, resp);
+            return;
+        }
+        if (driverLogin == null) {
+            resp.sendRedirect(req.getContextPath() + "/login");
+            return;
+        }
```
MaliukDaria, line 39: same, could we use only one if statement here?

MaliukDaria (changes requested):

PR #240, opened 2022-05-20
add solution fo jv-web-security

SerhiiPankov:


src/main/java/taxi/service/AuthenticationServiceImpl.java
```diff
@@ -0,0 +1,25 @@
...
+    
+    @Override
+    public Driver login(String login, String password) throws AuthenticationException {
+        Optional<Driver> driver = driverService.findByLogin(login);
+        if (driver.isEmpty()) {
+            throw new AuthenticationException("Login or password was incorrect...");
+        }
+        if (driver.get().getPassword().equals(password)) {
+            return driver.get();
+        }
```
MaliukDaria, line 22: could we use only one if statement here?

src/main/java/taxi/model/Driver.java
```diff
@@ -6,13 +6,33 @@ public class Driver {
     private Long id;
     private String name;
     private String licenseNumber;
+    private String login;
+    private String password;
 
     public Driver() {
```
MaliukDaria, line 12: I think we should override toString method here to show info about drivers in exception messages(but don't include password field)

src/main/java/taxi/web/filter/AuthenticationFilter.java
```diff
@@ -0,0 +1,42 @@
...
+        HttpSession session = httpServletRequest.getSession();
+        Long userId = (Long) session.getAttribute("driver_id");
+        if (userId == null && allowedUrl.contains(httpServletRequest.getServletPath())) {
+            chain.doFilter(httpServletRequest, httpServletResponse);
+            return;
+        }
+        if (userId == null) {
+            httpServletResponse.sendRedirect(httpServletRequest.getContextPath() + "/login");
+            return;
+        }
```
MaliukDaria, line 39: same? could we use only one if statement here?

MaliukDaria (changes requested):

src/main/java/taxi/web/filter/AuthenticationFilter.java
```diff
@@ -0,0 +1,38 @@
...
+            throws IOException, ServletException {
+        HttpServletRequest httpServletRequest = (HttpServletRequest) request;
+        HttpServletResponse httpServletResponse = (HttpServletResponse) response;
+        HttpSession session = httpServletRequest.getSession();
+        Long userId = (Long) session.getAttribute("driver_id");
+        if (userId != null || allowedUrl.contains(httpServletRequest.getServletPath())) {
+            chain.doFilter(httpServletRequest, httpServletResponse);
+        } else {
+            httpServletResponse.sendRedirect(httpServletRequest.getContextPath() + "/login");
+        }
```
MaliukDaria, line 36:
```suggestion
        if (userId != null || allowedUrl.contains(httpServletRequest.getServletPath())) {
            chain.doFilter(httpServletRequest, httpServletResponse);
            return;
        } 
            httpServletResponse.sendRedirect(httpServletRequest.getContextPath() + "/login");
        }
```

MaliukDaria (approved):

PR #268, opened 2022-06-14
implemented authentication filters

DLyudmila:


src/main/java/taxi/controller/LogoutController.java
```diff
@@ -0,0 +1,16 @@
...
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+public class LogoutController extends HttpServlet {
+    @Override
+    protected void doGet(HttpServletRequest req, HttpServletResponse resp)
+            throws ServletException, IOException {
+        req.getSession().invalidate();
+        resp.sendRedirect("/login");
```
MaliukDaria, line 14: If you use sendRedirect() method in your controllers, please pass "request.getContextPath()" + "/your-endpoint" as a parameter. Currently, the context path is empty, but if it is not, your code still should work.

src/main/java/taxi/model/Driver.java
```diff
@@ -6,6 +6,8 @@ public class Driver {
     private Long id;
```
MaliukDaria, line 6: lets add `toString()` method to Driver class, so we have appropriate driver description in exceptions messages

src/main/java/taxi/web/filter/AuthenticationFilter.java
```diff
@@ -0,0 +1,45 @@
...
+
+        Long driverId = (Long) session.getAttribute("driver_id");
+        if (allowedUrls.contains(req.getServletPath())) {
+            filterChain.doFilter(req, resp);
+            return;
+        }
+        if (driverId == null) {
+            resp.sendRedirect("/login");
+            return;
+        }
```
MaliukDaria, line 42: could we make it using only one if statement?

MaliukDaria (changes requested):

PR #279, opened 2022-07-01
Jv web security solution

bigmozes:
Implemented all necessary classes, methods and pages, according to project README.

### DIFF
--- a/src/main/java/taxi/model/Driver.java
+++ b/src/main/java/taxi/model/Driver.java
@@ -12,6 +12,11 @@ public class Driver {
     public Driver() {
     }
 
+    public Driver(String name, String licenseNumber) {
+        this.name = name;
+        this.licenseNumber = licenseNumber;
+    }
+
     public Driver(String name, String licenseNumber, String login, String password) {
         this.name = name;
         this.licenseNumber = licenseNumber;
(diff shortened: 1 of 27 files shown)


src/main/java/taxi/web/filter/AuthenticationFilter.java
```diff
@@ -0,0 +1,44 @@
...
+        HttpSession session = req.getSession();
+        Long userId = (Long) session.getAttribute("user_id");
+        if (userId == null && allowedUrls.contains(req.getServletPath())) {
+            filterChain.doFilter(req, resp);
+            return;
+        }
+        if (userId == null) {
+            resp.sendRedirect("/login");
+            return;
+        }
```
MaliukDaria, line 41: could we use only one if statement here

src/main/java/taxi/model/Driver.java
```diff
@@ -6,6 +6,8 @@ public class Driver {
     private Long id;
     private String name;
     private String licenseNumber;
+    private String login;
+    private String password;
 
     public Driver() {
```
MaliukDaria, line 12: we should add `toString()` method to Driver class to display driver info correctly in exception messages(but not add password field to it)

bigmozes: Ok, done, thought it was optional.

src/main/java/taxi/web/filter/AuthenticationFilter.java
bigmozes: Done.

MaliukDaria (approved):

PR #282, opened 2022-07-04
Implemented login and logout services

ChumakAndrey:


src/main/java/taxi/model/Driver.java
```diff
@@ -6,6 +6,8 @@ public class Driver {
     private Long id;
     private String name;
     private String licenseNumber;
+    private String login;
+    private String password;
```
gdpbalt, line 10: You don't create this fields in src/main/resources/init_db.sql
I can't run your application

gdpbalt (changes requested):

src/main/java/taxi/service/AuthenticationServiceImpl.java
```diff
@@ -0,0 +1,25 @@
...
+    @Override
+    public Driver login(String username, String password) throws AuthenticationException {
+        Optional<Driver> driver = driverService.findByLogin(username);
+        if (driver.isEmpty()) {
+            throw new AuthenticationException("User name or password was incorrect");
+        }
+        if (driver.get().getPassword().equals(password)) {
+            return driver.get();
+        }
+        throw new AuthenticationException("User name or password was incorrect");
```
KostyaNikiforov, line 23: Can we throw an exception only once?

src/main/java/taxi/web/filter/AuthenticationFilter.java
```diff
@@ -0,0 +1,44 @@
...
+        Long driverId = (Long) session.getAttribute("driver_id");
+        if (driverId == null && allowedUrls.contains(req.getServletPath())) {
+            filterChain.doFilter(req,resp);
+            return;
+        }
+        if (driverId == null) {
+            resp.sendRedirect("/login");
+            return;
+        }
+        filterChain.doFilter(req, resp);
```
KostyaNikiforov, line 42: Let's think about how we can get rid of one `if-block`

KostyaNikiforov (approved):
Good job:fire:

PR #290, opened 2022-07-13
feat: authentication, cars of a logged in driver

mikenikolov:


MaliukDaria (changes requested):

src/main/java/taxi/service/AuthenticationServiceImpl.java
```diff
@@ -0,0 +1,23 @@
...
+public class AuthenticationServiceImpl implements AuthenticationService {
+    @Inject
+    private DriverService driverService;
+
+    @Override
+    public Driver login(String login, String password) throws AuthenticationException {
+        Optional<Driver> driver = driverService.findByLogin(login);
+        if (driver.isPresent() && driver.get().getPassword().equals(password)) {
+            return driver.get();
+        } else {
```
MaliukDaria, line 19: do we need else after return?

src/main/java/taxi/filter/AuthenticationFilter.java
```diff
@@ -0,0 +1,43 @@
...
+        Long driverId = (Long) session.getAttribute("driver_id");
+        req.setAttribute("isLoggedIn", driverId != null);
+        if (driverId == null && !allowedUrls.contains(req.getServletPath())) {
+            resp.sendRedirect(req.getContextPath() + "/login");
+            return;
+        }
+        if (driverId != null && req.getServletPath().equals(req.getContextPath() + "/login")) {
+            resp.sendRedirect(req.getContextPath() + "/index");
+            return;
+        }
```
MaliukDaria, line 40: could we use only one if statement here?

MaliukDaria (changes requested):

MaliukDaria (approved):

PR #296, opened 2022-07-15
task done

Pavlo99-gif:


src/main/webapp/WEB-INF/views/drivers/all.jsp
```diff
@@ -27,6 +29,9 @@
             <td>
                 <c:out value="${driver.licenseNumber}"/>
             </td>
+            <td>
+                <c:out value="${driver.login}"/>
```
YaroslavLa, line 33: This is sensitive information, so I don`t recommend displaying it on this page

YaroslavLa (changes requested):

YaroslavLa (approved):
Good Solution!

KostyaNikiforov (approved):

PR #299, opened 2022-07-18
Authentication for drivers were implemented

olyaklipa:


src/main/webapp/web.xml
```diff
@@ -105,4 +114,31 @@
...
+    </servlet-mapping>
+
+    <filter>
+        <filter-name>authenticationFilter</filter-name>
+        <filter-class>taxi.web.filter.AuthenticationFilter</filter-class>
+    </filter>
+    <filter-mapping>
+        <filter-name>authenticationFilter</filter-name>
+        <url-pattern>/*</url-pattern>
+    </filter-mapping>
```
antonEEmus, line 143: I prefer to use annotations as WebController and WebFilter, but it's up to you)

antonEEmus (changes requested):

src/main/webapp/web.xml
olyaklipa: Good point! I also like the annotation option. It is just because I have been given the web.xml with nearly complete list of servlet/mappings. So, decided to use this option in the end. So, I will keep it as it is.

antonEEmus (approved):

liliia-ponomarenko (approved):
great job 🚀

PR #304, opened 2022-07-25
Third try

alexgash8:


src/main/java/taxi/controller/car/GetAllCarsController.java
```diff
@@ -3,13 +3,15 @@
...
 import javax.servlet.ServletException;
+import javax.servlet.annotation.WebServlet;
 import javax.servlet.http.HttpServlet;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 import taxi.lib.Injector;
 import taxi.model.Car;
 import taxi.service.CarService;
 
+@WebServlet(urlPatterns = {"/cars/all", "/cars", "/cars/"})
```
MaliukDaria, line 14:
```suggestion
@WebServlet(urlPatterns = {"/cars", "/cars/"})
```

image

src/main/java/taxi/model/Driver.java
```diff
@@ -40,21 +60,12 @@ public void setLicenseNumber(String licenseNumber) {
...
-        }
-        if (o == null || getClass() != o.getClass()) {
-            return false;
-        }
-        Driver driver = (Driver) o;
-        return Objects.equals(id, driver.id)
-                && Objects.equals(name, driver.name)
-                && Objects.equals(licenseNumber, driver.licenseNumber);
+    public boolean equals(Object obj) {
+        return super.equals(obj);
```
MaliukDaria, line 64: why do we need to change equals logic?

src/main/webapp/WEB-INF/views/drivers/add.jsp
```diff
@@ -23,10 +25,17 @@
         <td>
             <input type="text" name="license_number" form="driver" required>
         </td>
+        <td>
+            <input type="text" name="login" form="driver" required>
+        </td>
+        <td>
+            <input type="text" name="password" form="driver" required>
```
MaliukDaria, line 32:
```suggestion
            <input type="password" name="password" form="driver" required>
```

MaliukDaria (changes requested):

src/main/java/taxi/model/Driver.java
alexgash8: because we added to Driver two more fields? Remade implementation...

MaliukDaria (approved):

PR #323, opened 2022-08-30
created login and logout controllers, implemented Authentication Service and all required logic

ssplot56:


src/main/java/taxi/dao/DriverDaoImpl.java
```diff
@@ -101,14 +106,36 @@ public boolean delete(Long id) {
         }
     }
 
+    @Override
+    public Optional<Driver> findByLogin(String login) {
+        String query = "SELECT * FROM drivers WHERE login = ? AND is_deleted = false";
```
Nick97-git, line 111:
```suggestion
        String query = "SELECT * FROM drivers WHERE login = ? AND is_deleted = FALSE";
```

Nick97-git (approved):

PR #328, opened 2022-09-04
completed all tasks

vitaliy1210:


src/main/java/taxi/model/Driver.java
```diff
@@ -4,15 +4,19 @@
 
 public class Driver {
```
MaliukDaria, line 5: It's better to add `toString()` method here, to get more informative exception messages in dao layer (but do not include password field there)

src/main/java/taxi/service/AuthenticationServiceImpl.java
```diff
@@ -0,0 +1,22 @@
...
+import java.util.Optional;
+import taxi.exception.AuthenticationException;
+import taxi.lib.Injector;
+import taxi.lib.Service;
+import taxi.model.Driver;
+
+@Service
+public class AuthenticationServiceImpl implements AuthenticationService {
+    private static final Injector injector = Injector.getInstance("taxi");
+    private DriverService driverService = (DriverService) injector.getInstance(DriverService.class);
```
MaliukDaria, line 12: we should use @Inject annotation here

src/main/java/taxi/service/DriverServiceImpl.java
```diff
@@ -18,7 +19,7 @@ public Driver create(Driver driver) {
 
     @Override
     public Driver get(Long id) {
-        return driverDao.get(id).get();
+        return driverDao.get(id).orElseGet(null);
```
MaliukDaria, line 22: why we need to get null Driver here?

src/main/java/taxi/web/filter/AuthenticationFilter.java
```diff
@@ -0,0 +1,44 @@
...
+        HttpSession session = req.getSession();
+        Long userId = (Long) session.getAttribute("user_id");
+        if (allowedUrls.contains(req.getServletPath())) {
+            filterChain.doFilter(req, resp);
+            return;
+        }
+        if (userId == null) {
+            resp.sendRedirect(req.getContextPath() + "/login");
+            return;
+        }
```
MaliukDaria, line 41: could we use only one if statement here?

MaliukDaria (changes requested):

zachar-chorny (approved):

PR #338, opened 2022-09-20
all the required classes and methods implemented, jsp files created

max-bolotin:


Maksym-Korkosh (approved):

Noct2000 (changes requested):

src/main/java/taxi/controller/driver/AddDriverController.java
```diff
@@ -24,8 +24,10 @@ public void doGet(HttpServletRequest req, HttpServletResponse resp)
     public void doPost(HttpServletRequest req, HttpServletResponse resp) throws IOException {
         String name = req.getParameter("name");
         String licenseNumber = req.getParameter("license_number");
-        Driver driver = new Driver(name, licenseNumber);
+        String login = req.getParameter("login");
+        String password = req.getParameter("password");
+        Driver driver = new Driver(name, licenseNumber, login, password);
         driverService.create(driver);
-        resp.sendRedirect(req.getContextPath() + "/drivers/add");
+        resp.sendRedirect(req.getContextPath() + "/index");
```
Noct2000, line 31: now "/index" is allowed only for authenticated users, let's revert changes
```suggestion
        resp.sendRedirect(req.getContextPath() + "/drivers/add");
```

Noct2000 (changes requested):

src/main/java/taxi/controller/driver/AddDriverController.java
max-bolotin: I disagree with this. By applying these changes, the user will stay at the Register page. I believe this is wrong behaviour. In my implementation with redirect to resp.sendRedirect(req.getContextPath() + "/index") the user after registration will get to the login page again. And after logging in, he'll get to the index page. The index page must not be available for unauthorised users according to the task.
I prefer to change the page **not** to "/drivers/add" but to "/login" to avoid double redirecting.

Noct2000 (approved):

PR #366, opened 2022-09-22
web security is completed

Kupriiuk:


JustElvis (approved):

src/main/java/taxi/model/Driver.java
```diff
@@ -48,13 +68,23 @@ public boolean equals(Object o) {
...
+        return Objects.hash(id, name, licenseNumber, login, password);
+    }
+
+    @Override
+    public String toString() {
+        return "Driver{" + "id=" + id
+                + ", name='" + name + '\''
+                + ", licenseNumber='" + licenseNumber + '\''
+                + ", login='" + login + '\''
+                + ", password='" + password + '\'' + '}';
```
Sofasmile, line 88: let's remove password from toString method

Sofasmile (changes requested):

PR #378, opened 2022-09-22
taxi service implementation was done

not4und:


src/main/java/taxi/web/filter/AuthenticationFilter.java
```diff
@@ -0,0 +1,42 @@
...
+        HttpSession session = req.getSession();
+        Long driverId = (Long) session.getAttribute("driver_id");
+        if (driverId == null && allowedUrls.contains(req.getServletPath())) {
+            filterChain.doFilter(req, resp);
+            return;
+        }
+        if (driverId == null) {
+            resp.sendRedirect(req.getContextPath() + "/login");
+            return;
+        }
```
aaananas, line 39: lets refactor it, lets make it just 1 if
```suggestion
        if (driverId == null && !allowedUrls.contains(servletPath)) {
            resp.sendRedirect("/login");
            return;
        }
```

aaananas (changes requested):

kozhukhovsky (approved):

PR #409, opened 2022-11-17
implemented task

vov4ik89:


src/main/java/taxi/service/AuthenticationServiceImpl.java
```diff
@@ -0,0 +1,19 @@
+package taxi.service;
+
+import java.util.Optional;
+import taxi.dao.DriverDao;
+import taxi.exception.AuthenticationException;
+import taxi.model.Driver;
+
+public class AuthenticationServiceImpl implements AuthenticationService {
+    private DriverDao driverDao;
```
SoulARC, line 9: It is better to use DriverService and @.Inject

src/main/java/taxi/service/AuthenticationServiceImpl.java
```diff
@@ -0,0 +1,19 @@
+package taxi.service;
+
+import java.util.Optional;
+import taxi.dao.DriverDao;
+import taxi.exception.AuthenticationException;
+import taxi.model.Driver;
+
```
SoulARC, line 7: @.Service

SoulARC (changes requested):

SoulARC (approved):
LGTM!

PR #414, opened 2022-11-21
Finished all tasks from web-security assignment

igorgalenko:


src/main/java/taxi/web/filter/AuthenticationFilter.java
```diff
@@ -0,0 +1,42 @@
...
+        HttpSession session = req.getSession();
+        Long driverId = (Long) session.getAttribute("driver_id");
+        if (driverId == null && allowedLinks.contains(req.getServletPath())) {
+            chain.doFilter(request, response);
+            return;
+        }
+        if (driverId == null) {
+            resp.sendRedirect("/login");
+            return;
+        }
```
YuriiHryniv, line 39: You can avoid using two if statements by placing the second one above the first one and modyfing it with

YuriiHryniv (approved):
LGTM

src/main/resources/init_db.sql
```diff
@@ -56,3 +56,7 @@ CREATE TABLE `cars_drivers`  (
 ) ENGINE = InnoDB CHARACTER SET = utf8 COLLATE = utf8_general_ci ROW_FORMAT = Dynamic;
 
 SET FOREIGN_KEY_CHECKS = 1;
+
+alter table drivers
+    add login varchar(50) null,
+    add password varchar(11) null;
```
liliia-ponomarenko, line 62: better add these columns to CREATE TABLE `drivers` script 
Also, you have to validate these parameters (login varchar(50), password varchar(11)) on the API side because you haven't the same constraints in your ValidationService and vice versa you validate that login and password  can't be null but don’t have the same constraints  here
this can cause unexpected exceptions to occur

liliia-ponomarenko (changes requested):
great! Let’s fix a few mistakes 😉

liliia-ponomarenko (approved):
great🔥

PR #423, opened 2022-11-25
Add three fields to Driver model - login, salt, password;Modified the…

PavloPolovyi:
… required JSP pages and corresponding controllers to support two new fields. Modified DriverDaoImpl and CarDaoImpl to fetch all columns for drivers, including new ones. Created new AuthenticationService with method Driver login(String login, String password); Created PasswordManager to hold operation with password - hashing, getting salt and checking, created RegistrationService to hold creation of new Driver, created AuthenticationFilter to filter those users, who are not login. Created GetMyCurrentCarsController for showing all cars that current driver has - used for this the same jsp page, that is used for showing all cars. Created LoginController and LogoutController.

### DIFF
--- a/src/main/java/taxi/controller/driver/GetMyCurrentCarsController.java
+++ b/src/main/java/taxi/controller/driver/GetMyCurrentCarsController.java
@@ -1,0 +1,26 @@
+package taxi.controller.driver;
+
+import java.io.IOException;
+import java.util.List;
+import javax.servlet.ServletException;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import taxi.lib.Injector;
+import taxi.model.Car;
+import taxi.service.CarService;
+
+public class GetMyCurrentCarsController extends HttpServlet {
+    private static final Injector injector = Injector.getInstance("taxi");
+    private final CarService carService =
+            (CarService) injector.getInstance(CarService.class);
+
+    @Override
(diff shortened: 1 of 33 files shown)


src/main/java/taxi/controller/driver/GetMyCurrentCarsController.java
```diff
@@ -0,0 +1,26 @@
...
+    private final CarService carService =
+            (CarService) injector.getInstance(CarService.class);
+
+    @Override
+    protected void doGet(HttpServletRequest req, HttpServletResponse resp)
+            throws ServletException, IOException {
+        Long driverId = (Long) req.getSession().getAttribute("driver_id");
+        List<Car> allCars = carService.getAllByDriver(driverId);
+        req.setAttribute("cars", allCars);
+        req.getRequestDispatcher("/WEB-INF/views/cars/all.jsp").forward(req, resp);
```
muzykaVP, line 24: i think our end point name shoud contain driver cars, not just cars,
something like this: /driver/cars
so those cars belong to current driver, not cars in general

muzykaVP (commented):
LGTM 👍 
one minor remark

hronograf (approved):
Really good! 🍉🍉

src/main/java/taxi/controller/driver/GetMyCurrentCarsController.java
PavloPolovyi: sorry for answering so late, but endpoint is 'drivers/cars', i just use the same jsp page ( "/WEB-INF/views/cars/all.jsp") for showing all cars for driver as for endpoint '/cars'.

PR #496, opened 2023-02-16
Adds implemented login, logout and get cars for logged driver pages

BusyhinD:


src/main/java/taxi/controller/LogoutController.java
```diff
@@ -0,0 +1,17 @@
+package taxi.controller;
+
+import java.io.IOException;
+import javax.servlet.ServletException;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+public class LogoutController extends HttpServlet {
+
```
Ivan95kos, line 10: For the future, remove the first empty line in all classes/interfaces/methods where it is present

Ivan95kos (approved):
Great job

src/main/java/taxi/controller/LogoutController.java
BusyhinD: Thank you! I'll remember it.

PR #500, opened 2023-02-21
First commit

Taurusden:
1. Fields password and login were added to class driver.
2. All methods which have connection with class Driver were changed.
3. Created two controllers for login and authentication.
4. Created redirections for logged and unlogged users.

### DIFF
--- a/src/main/java/taxi/controller/driver/LoginController.java
+++ b/src/main/java/taxi/controller/driver/LoginController.java
@@ -1,6 +1,7 @@
 package taxi.controller.driver;
 
 import java.io.IOException;
+import javax.security.sasl.AuthenticationException;
 import javax.servlet.ServletException;
 import javax.servlet.annotation.WebServlet;
 import javax.servlet.http.HttpServlet;
@@ -34,7 +35,7 @@ public class LoginController extends HttpServlet {
             HttpSession session = req.getSession();
             session.setAttribute("driver_id", driver.getId());
             resp.sendRedirect("/cars");
-        } catch (RuntimeException e) {
+        } catch (AuthenticationException e) {
             req.setAttribute("errorMessage", e.getMessage());
             req.getRequestDispatcher("/WEB-INF/views/login.jsp").forward(req, resp);
         }
--- a/src/main/java/taxi/controller/driver/LogoutController.java
+++ b/src/main/java/taxi/controller/driver/LogoutController.java
@@ -1,8 +1,19 @@
 package taxi.controller.driver;
 
+import java.io.IOException;
+import javax.servlet.ServletException;
 import javax.servlet.annotation.WebServlet;
 import javax.servlet.http.HttpServlet;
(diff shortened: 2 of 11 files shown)


src/main/java/taxi/controller/driver/LogoutController.java
```diff
@@ -0,0 +1,8 @@
+package taxi.controller.driver;
+
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+
+@WebServlet(urlPatterns = "/drivers/logout")
+public class LogoutController extends HttpServlet {
+}
```
olekskov, line 8: Is logout controller works without any logic inside?

src/main/java/taxi/model/Driver.java
```diff
@@ -50,11 +75,21 @@ public boolean equals(Object o) {
         Driver driver = (Driver) o;
         return Objects.equals(id, driver.id)
                 && Objects.equals(name, driver.name)
-                && Objects.equals(licenseNumber, driver.licenseNumber);
+                && Objects.equals(licenseNumber, driver.licenseNumber)
+                && Objects.equals(login, driver.login)
+                && Objects.equals(password, driver.password);
```
olekskov, line 80: It's not a good practice to use user credentials anywhere outside of places where it's necessary for authorisation or authentication process

src/main/java/taxi/controller/driver/LoginController.java
```diff
@@ -0,0 +1,43 @@
...
+            throws ServletException, IOException {
+        String login = req.getParameter("login");
+        String password = req.getParameter("password");
+
+        try {
+            Driver driver = authenticationService.login(login, password);
+            HttpSession session = req.getSession();
+            session.setAttribute("driver_id", driver.getId());
+            resp.sendRedirect("/cars");
+        } catch (RuntimeException e) {
```
olekskov, line 37: Type of exception should be more precise

src/main/java/taxi/service/AuthenticationServiceImpl.java
```diff
@@ -0,0 +1,22 @@
...
+@Service
+public class AuthenticationServiceImpl implements AuthenticationService {
+    @Inject
+    private DriverService driverService;
+
+    @Override
+    public Driver login(String login, String password) {
+        Driver driver = driverService.findByLogin(login)
+                .orElseThrow(() ->
+                        new RuntimeException("Login or password was incorrect"));
```
olekskov, line 16: Please chose more narrow type of exception

olekskov (changes requested):

src/main/java/taxi/controller/driver/LogoutController.java
Taurusden: Created with login and forgot, cause in task it is not needed).

Ivan95kos (approved):
Great job

PR #502, opened 2023-02-21
HW is done. Waiting for the review.

Bogdan-Pokhilko:


src/main/java/taxi/dao/CarDaoImpl.java
```diff
@@ -17,18 +18,19 @@
 
 @Dao
 public class CarDaoImpl implements CarDao {
```
Rommelua, line 20: revert all unnecessary changes in this class

Rommelua (changes requested):

aaananas (approved):

PR #509, opened 2023-02-24
feat: jv-web-security #comment added login page

sania517:


src/main/java/taxi/model/Driver.java
```diff
@@ -15,6 +17,13 @@ public Driver(String name, String licenseNumber) {
         this.licenseNumber = licenseNumber;
     }
 
+    public Driver(String name, String licenseNumber, String login, String password) {
+        this.name = name;
+        this.licenseNumber = licenseNumber;
```
olekskov, line 22: Please use constructor above to set part of properties here

src/main/java/taxi/dao/DriverDaoImpl.java
```diff
@@ -98,14 +102,35 @@ public boolean delete(Long id) {
         }
     }
 
+    @Override
+    public Optional<Driver> findByLogin(String login) {
+        String query = "SELECT * FROM drivers WHERE login = ?";
```
olekskov, line 107: should we be able to find deleted driver?

olekskov (changes requested):

Ivan95kos (approved):
Great job

kozhukhovsky (approved):

PR #515, opened 2023-03-04
jv-web-security solution

nkharyshyn:
-added fields login and password to my class Driver and to driver table in my db 
-modified existing methods and JSP pages as needed after adding login and password for drivers 
-created  AuthenticationService with method login
-created method findByLogin in DriverDao and DriverService and used it to identify the user based on the login in AuthenticationService 
-implemented LoginController and LogoutController with corresponding JSP pages 
-created AuthenticationFilter in such way that only LoginController and CreateDriverController should be available for non-authenticated users 
-created GetMyCurrentCarsController to display all car for currently logged driver

### DIFF
--- a/src/main/java/taxi/controller/driver/AddDriverController.java
+++ b/src/main/java/taxi/controller/driver/AddDriverController.java
@@ -21,10 +21,13 @@ public class AddDriverController extends HttpServlet {
     }
 
     @Override
-    public void doPost(HttpServletRequest req, HttpServletResponse resp) throws IOException {
+    public void doPost(HttpServletRequest req, HttpServletResponse resp)
+            throws IOException {
         String name = req.getParameter("name");
         String licenseNumber = req.getParameter("license_number");
-        Driver driver = new Driver(name, licenseNumber);
+        String login = req.getParameter("login");
+        String password = req.getParameter("password");
+        Driver driver = new Driver(name, licenseNumber, login, password);
         driverService.create(driver);
         resp.sendRedirect(req.getContextPath() + "/drivers/add");
     }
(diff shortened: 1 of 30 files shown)


src/main/java/taxi/controller/driver/AddDriverController.java
```diff
@@ -21,10 +21,13 @@ public void doGet(HttpServletRequest req, HttpServletResponse resp)
     }
 
     @Override
-    public void doPost(HttpServletRequest req, HttpServletResponse resp) throws IOException {
+    public void doPost(HttpServletRequest req, HttpServletResponse resp)
+            throws IOException {
```
olekskov, line 25: what is the reason to split method header into 2 lines?

src/main/java/taxi/model/Driver.java
```diff
@@ -50,11 +70,13 @@ public boolean equals(Object o) {
...
-                && Objects.equals(licenseNumber, driver.licenseNumber);
+                && Objects.equals(licenseNumber, driver.licenseNumber)
+                && Objects.equals(login, driver.login)
+                && Objects.equals(password, driver.password);
     }
 
     @Override
     public int hashCode() {
-        return Objects.hash(id, name, licenseNumber);
+        return Objects.hash(id, name, licenseNumber, login, password);
```
olekskov, line 80: Please do not use confidential data anywhere outside absolutely necessary logic for authentication

olekskov (changes requested):

uliana-andrushchuk (changes requested):
Previous comments not fixed

olekskov (approved):
Good job!

PR #525, opened 2023-03-20
Implemented Authentication

mishok321:


src/main/java/taxi/controller/LoginController.java
```diff
@@ -0,0 +1,46 @@
...
+            return;
+        }
+        req.getRequestDispatcher("/WEB-INF/views/login.jsp").forward(req, resp);
+    }
+
+    @Override
+    public void doPost(HttpServletRequest req, HttpServletResponse resp)
+            throws ServletException, IOException {
+        AuthenticationService authenticationService
+                = (AuthenticationService) injector.getInstance(AuthenticationService.class);
```
nacenik, line 33: Better to create services as class-level variables, cause you can reuse them in the future.

nacenik (approved):
LGTM! 👍

PR #527, opened 2023-03-22
implemented login and logout controllers, added jsp for them, added f…

YevheniiKilovyi:
…ields login and password for Driver class, made all DAOs and services match with these fields, implemented AuthService, added costum Authentication exception and implemented controller for getting cars for current user


Ivan95kos (approved):
Great job

PR #535, opened 2023-03-28
test

Watxesnavo:
TEST error with deploying on server.
some problem with filter class


src/main/java/taxi/controller/LoginController.java
```diff
@@ -0,0 +1,40 @@
...
+    @Override
+    protected void doPost(HttpServletRequest req, HttpServletResponse resp)
+            throws ServletException, IOException {
+        String userName = req.getParameter("login");
+        String password = req.getParameter("password");
+        try {
+            Driver driver = authenticationService.login(userName, password);
+            HttpSession session = req.getSession();
+            session.setAttribute("driver_id", driver.getId());
+            resp.sendRedirect("/index");
```
kozhukhovsky, line 34: I doubt this will work correctly without req.getContextPath()

Fix it in other place too

kozhukhovsky (changes requested):

kozhukhovsky (approved):

PR #548, opened 2023-04-07
Servlet Web Security completed

IlkoGood:


src/main/java/taxi/model/Driver.java
```diff
@@ -39,6 +41,22 @@ public void setLicenseNumber(String licenseNumber) {
...
+    }
+
+    public String getPassword() {
+        return password;
+    }
+
+    public void setPassword(String password) {
+        this.password = password;
+    }
+
```
sarakhmen, line 59: I'd also recommend adding new fields to the equals and hashcode methods.

src/main/java/taxi/dao/DriverDaoImpl.java
```diff
@@ -17,13 +17,15 @@
 public class DriverDaoImpl implements DriverDao {
```
sarakhmen, line 17: Let's also move your implementations to the impl folder

sarakhmen (changes requested):
Let's improve your solution :)

src/main/java/taxi/model/Driver.java
IlkoGood: we can easily remove that methods... but ok

src/main/java/taxi/dao/DriverDaoImpl.java
IlkoGood: Just left it as original... but ok again)

nacenik (approved):
LGTM! 👍

PR #554, opened 2023-04-09
added homework

sanya-1xx8:


src/main/java/taxi/controller/driver/GetMyCurrentCarsController.java
```diff
@@ -0,0 +1,25 @@
...
+import taxi.service.CarService;
+
+public class GetMyCurrentCarsController extends HttpServlet {
+    private static final Injector injector = Injector.getInstance("taxi");
+    private final CarService carService = (CarService) injector.getInstance(CarService.class);
+
+    @Override
+    protected void doGet(HttpServletRequest req, HttpServletResponse resp)
+            throws ServletException, IOException {
+        Long driverId = Long.valueOf(req.getSession().getAttribute("id").toString());
```
AlexandLes, line 20: you can cast just using (Long) and without toString method

AlexandLes (approved):
Well Done!

src/main/java/taxi/service/AuthenticationServiceImpl.java
```diff
@@ -0,0 +1,34 @@
...
+    public Driver register(String username, String password, String repeatPassword)
+            throws AuthenticationException {
+        if (password.equals(repeatPassword)) {
+            Driver driver = new Driver();
+            driver.setLogin(username);
+            driver.setPassword(password);
+            return driver;
+        }
+        throw new AuthenticationException("Passwords don't match, please try again!");
+    }
```
petro-ushchuk, line 33: is it possible to register driver without license number and name?

src/main/resources/init_db.sql
```diff
@@ -11,6 +11,8 @@ CREATE TABLE `drivers`  (
                             `id` BIGINT(0) UNSIGNED NOT NULL AUTO_INCREMENT,
                             `name` VARCHAR(255) CHARACTER SET utf8 COLLATE utf8_general_ci NOT NULL,
                             `license_number` VARCHAR(255) CHARACTER SET utf8 COLLATE utf8_general_ci NOT NULL,
+                            `login` VARCHAR(255) CHARACTER SET utf8 COLLATE utf8_general_ci NOT NULL,
```
petro-ushchuk, line 14: login should be unique, isn't it?

petro-ushchuk (changes requested):
Good work, but we should make it better 🔥

src/main/resources/init_db.sql
sanya-1xx8: I think so)

sarakhmen (approved):
Good job ;)

PR #587, opened 2023-04-29
Solved

arthursilver111:


src/main/java/taxi/web/filter/AuthenticationFilter.java
```diff
@@ -0,0 +1,40 @@
...
+        HttpServletRequest httpServletRequest = (HttpServletRequest) servletRequest;
+        HttpServletResponse httpServletResponse = (HttpServletResponse) servletResponse;
+        HttpSession session = httpServletRequest.getSession();
+        Long driverId = (Long) session.getAttribute("driver_id");
+
+        if (driverId == null && allowedUrl.contains(httpServletRequest.getServletPath())) {
+            filterChain.doFilter(servletRequest, servletResponse);
+            return;
+        }
+        filterChain.doFilter(servletRequest, servletResponse);
```
olekskov, line 38: There is no sense in this code right now:
The condition should check that driverId is not null or the URL is allowed. And redirect to login page or perform doFilter() method.

olekskov (changes requested):

src/main/java/taxi/controller/authentication/LoginController.java
```diff
@@ -0,0 +1,40 @@
...
+    @Override
+    protected void doPost(HttpServletRequest req, HttpServletResponse resp)
+            throws ServletException, IOException {
+        String login = req.getParameter("login");
+        String password = req.getParameter("password");
+        try {
+            Driver driver = authenticationService.login(login, password);
+            HttpSession session = req.getSession();
+            session.setAttribute("driver_id", driver.getId());
+            resp.sendRedirect("/index");
```
olekskov, line 34: contextPath not fixed

src/main/java/taxi/web/filter/AuthenticationFilter.java
```diff
@@ -0,0 +1,43 @@
...
+        Long driverId = (Long) session.getAttribute("driver_id");
+
+        if (driverId == null && allowedUrl.contains(httpServletRequest.getServletPath())) {
+            filterChain.doFilter(servletRequest, servletResponse);
+            return;
+        }
+        if (driverId == null) {
+            httpServletResponse.sendRedirect(httpServletRequest.getContextPath() + "/login");
+        }
+        filterChain.doFilter(servletRequest, servletResponse);
```
olekskov, line 41:
```suggestion
        if (driverId == null || !allowedUrl.contains(httpServletRequest.getServletPath())) {
            httpServletResponse.sendRedirect(httpServletRequest.getContextPath() + "/login");
        }
        filterChain.doFilter(servletRequest, servletResponse);
```

olekskov (changes requested):

src/main/java/taxi/web/filter/AuthenticationFilter.java
arthursilver111: In last version i used wrote

 if (driverId == null && !allowedUrl.contains(req.getServletPath())) {
            resp.sendRedirect(req.getContextPath() + "/login");
        }

becouse with || authentication does not work

olekskov: Yes, you're correct in latest version, it should be &&

olekskov (approved):
Good job

PR #594, opened 2023-05-20
most of the tasks done

ItsUnbidden:


src/main/java/taxi/web/filter/AuthenticationFilter.java
```diff
@@ -0,0 +1,60 @@
...
+                req.getSession().invalidate();
+                resp.sendRedirect(req.getContextPath() + "/login");
+                return;
+            }
+        }
+        // This was not a part of the task, so this 'if' doesn't count.
+        if (id != null && req.getServletPath().equals("/login")) {
+            resp.sendRedirect(req.getContextPath() + "/index");
+            return;
+        }
```
andriy-shulepa, line 57: what do you mean 'doesn't count'? if you wanted to write some custom logic - it's ok, but explain what are you doing here. if you don't want mentor to see some extra code - just don't commit it

src/main/java/taxi/model/Driver.java
```diff
@@ -39,22 +45,81 @@ public void setLicenseNumber(String licenseNumber) {
...
+    public int hashCode() {
+        final int prime = 31;
+        int result = 1;
+        result = prime * result + ((id == null) ? 0 : id.hashCode());
+        result = prime * result + ((name == null) ? 0 : name.hashCode());
+        result = prime * result + ((licenseNumber == null) ? 0 : licenseNumber.hashCode());
+        result = prime * result + ((login == null) ? 0 : login.hashCode());
+        result = prime * result + ((password == null) ? 0 : password.hashCode());
+        return result;
+    }
```
andriy-shulepa, line 74: you can add new fields to the existing approach instead of changing the approach. It will need fewer changes that you made

andriy-shulepa (changes requested):

src/main/java/taxi/web/filter/AuthenticationFilter.java
```diff
@@ -0,0 +1,60 @@
...
+        // Not sure whether this was requested in README
+        if (id != null) {
+            try {
+                driverService.get(id);
+            } catch (NoSuchElementException e) {
+                req.getSession().invalidate();
+                resp.sendRedirect(req.getContextPath() + "/login");
+                return;
+            }
+        }
```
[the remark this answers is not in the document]
ItsUnbidden: This is not the reason why I added this code. If a driver gets deleted while logged in, they will still be able to access all of the pages. To avoid this, I decided to check whether the id of this driver still exists in the db.

src/main/java/taxi/controller/driver/AddDriverController.java
```diff
@@ -17,14 +17,21 @@ public class AddDriverController extends HttpServlet {
     @Override
     public void doGet(HttpServletRequest req, HttpServletResponse resp)
             throws ServletException, IOException {
+        // This was not a part of the task. It's required for the button in the header 
+        // to change depending on whether the driver is already logged in.
+        if (req.getSession().getAttribute("driver_id") != null) {
+            req.setAttribute("is_logged_in", true);
+        }
```
Rommelua, line 24: Usually, registration and login are separate actions. After u registered, u should login.

src/main/java/taxi/service/DriverService.java
```diff
@@ -3,4 +3,5 @@
 import taxi.model.Driver;
 
 public interface DriverService extends GenericService<Driver> {
+    Driver findByLogin(String login);
```
Rommelua, line 6: image

src/main/java/taxi/service/AuthenticationServiceImpl.java
```diff
@@ -0,0 +1,27 @@
...
+    public Driver login(String login, String password) throws AuthenticationException {
+        Driver driver = null;
+        try {
+            driver = driverService.findByLogin(login);
+        } catch (NoSuchElementException e) {
+            throw new AuthenticationException("Login or password is invalid");
+        }
+        if (driver.getPassword().equals(password)) {
+            return driver;
+        }
```
Rommelua, line 24:
```suggestion
        Optional<Driver> driver = driverService.findByLogin(login);
        if (driver.isPresent() && driver.getPassword().equals(password)) {
            return driver.get();
        }
```

src/main/java/taxi/web/filter/AuthenticationFilter.java
```diff
@@ -0,0 +1,63 @@
...
+        HttpSession session = req.getSession();
+        Long id = (Long) session.getAttribute("driver_id");
+
+        if (id == null && !allowedUrls.contains(req.getServletPath())) {
+            resp.sendRedirect(req.getContextPath() + "/login");
+            return;
+        }
+        // Checks whether this driver's id still exists in the db. 
+        // Otherwise there is a possibility that a driver, 
+        // which has already been deleted, might be able to access the pages.
```
Rommelua, line 45: The best practice is to create code that is clear without comments (self-documented code). 
You can use private methods with meaningful names for them.

src/main/java/taxi/web/filter/AuthenticationFilter.java
```diff
@@ -0,0 +1,63 @@
...
+        if (id != null) {
+            try {
+                driverService.get(id);
+            } catch (NoSuchElementException e) {
+                req.getSession().invalidate();
+                resp.sendRedirect(req.getContextPath() + "/login");
+                return;
+            }
+        }
```
Rommelua, line 54:
```suggestion
        if (id == null || isDriverDeleted(id)) {
            req.getSession().invalidate();
            resp.sendRedirect(req.getContextPath() + "/login");
            return;
        }
```

Rommelua (approved):

PR #595, opened 2023-05-20
added proper solution for this task

polikarpov13:


src/main/java/taxi/web/filter/AuthenticationFilter.java
```diff
@@ -0,0 +1,46 @@
...
+        HttpSession httpSession = httpServletRequest.getSession();
+        if (httpSession.getAttribute("driver_id") == null
+                && allowedUrl.contains(httpServletRequest.getServletPath())) {
+            filterChain.doFilter(httpServletRequest, httpServletResponse);
+            return;
+        }
+        if (httpSession.getAttribute("driver_id") == null) {
+            httpServletResponse.sendRedirect("/login");
+            return;
+        }
```
olekskov, line 43: may be simplified
```suggestion
        if (httpSession.getAttribute("driver_id") == null
                && !allowedUrl.contains(httpServletRequest.getServletPath())) {
            httpServletResponse.sendRedirect("/login");
            return;
        }

```

olekskov (changes requested):

aaananas (approved):

PR #620, opened 2023-06-21
My branch

V-Solianyk:


src/main/java/taxi/service/DriverService.java
```diff
@@ -3,4 +3,5 @@
 import taxi.model.Driver;
 
 public interface DriverService extends GenericService<Driver> {
+    Driver findByLogin(String login);
```
marimarig, line 6:
```suggestion
    Optional<Driver> findByLogin(String login);
}
```

src/main/java/taxi/service/DriverServiceImpl.java
```diff
@@ -38,4 +38,10 @@ public Driver update(Driver driver) {
     public boolean delete(Long id) {
         return driverDao.delete(id);
     }
+
+    @Override
+    public Driver findByLogin(String login) {
```
marimarig, line 43: the same: public Optional<Driver> findByLogin(String login)

src/main/java/taxi/service/AuthenticationServiceImpl.java
```diff
@@ -0,0 +1,26 @@
...
+    @Override
+    public Driver login(String login, String password) throws AuthenticationException {
+        Optional<Driver> driver = driverDao.findByLogin(login);
+        if (driver.isEmpty()) {
+            throw new AuthenticationException("Name of driver or password was incorrect");
+        }
+        if (driver.get().getPassword().equals(password)) {
+            return driver.get();
+        }
+        throw new AuthenticationException("Name of driver or password was incorrect");
```
marimarig, line 24: in AuthenticationServiceImp try to use only one if block

marimarig (changes requested):
Good job, let's improve it a little bit)

src/main/java/taxi/service/DriverService.java
V-Solianyk: In this case, I thought that returning the Optional at the service level would not be okay, so I did it this way, but at the DAO level I return the optional, so maybe we should wait for the mentors' comments?

src/main/java/taxi/service/DriverServiceImpl.java
V-Solianyk: the same situation)

marimarig (approved):
Good;)

src/main/java/taxi/dao/DriverDaoImpl.java
```diff
@@ -98,14 +102,36 @@ public boolean delete(Long id) {
...
         Driver driver = new Driver();
         driver.setId(id);
         driver.setName(name);
         driver.setLicenseNumber(licenseNumber);
+        driver.setLogin(login);
+        driver.setPassword(password);
         return driver;
     }
 }
+
```
sarakhmen, line 137:
```suggestion
```

sarakhmen (approved):
Good job ;)

PR #621, opened 2023-06-22
home-work/jv-web-security

andrii-subash:


src/main/java/taxi/service/RegistrationServiceImp.java
```diff
@@ -0,0 +1,32 @@
+package taxi.service;
+
+import taxi.dao.DriverDao;
+import taxi.exception.RegistrationException;
+import taxi.lib.Inject;
+import taxi.lib.Service;
+
+@Service
+public class RegistrationServiceImp implements RegistrationService {
```
overpathz, line 9: Just to understand more clearly.
What was this service designed for?

overpathz (changes requested):
Good job. Need to clarify some points.

src/main/java/taxi/service/RegistrationServiceImp.java
andrii-subash: For registration a new driver, This is my addition, which is not provided by the terms of the task. This service is to register a new driver, to check if there is a driver with the same login, and to check if the repeated password matchesc

src/main/webapp/web.xml
```diff
@@ -3,106 +3,12 @@
          xsi:schemaLocation="http://xmlns.jcp.org/xml/ns/javaee
                       http://xmlns.jcp.org/xml/ns/javaee/web-app_4_0.xsd"
          version="4.0">
-    <servlet>
-        <servlet-name>index</servlet-name>
-        <servlet-class>taxi.controller.IndexController</servlet-class>
-    </servlet>
-    <servlet-mapping>
-        <servlet-name>index</servlet-name>
```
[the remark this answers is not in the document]
andrii-subash: image
Because I get a "page not found" error when I use mapping, . When I use annotation everything works fine for me. I don't know why this happens

src/main/java/taxi/service/RegistrationServiceImp.java
andrii-subash: I think this is an important point in the registration process, which should not be considered a mistake

aaananas (approved):

PR #623, opened 2023-06-23
Implemented task. Add all needed Controllers and JSP pages

ykhmaruk:


src/main/webapp/WEB-INF/views/drivers/all.jsp
```diff
@@ -27,6 +29,9 @@
             <td>
                 <c:out value="${driver.licenseNumber}"/>
             </td>
+            <td>
+                <c:out value="${driver.login}"/>
+            </td>
```
viprko, line 34: Do we really need to display the driver's username on the list of all drivers?

viprko (approved):
Keep it up!

rovein (approved):
Great job!

PR #626, opened 2023-06-25
HW-jv-web-security-25-Of-June

ViktoriiaSvidovska:


src/main/java/taxi/service/DriverService.java
```diff
@@ -1,6 +1,8 @@
 package taxi.service;
 
+import taxi.exception.AuthenticationException;
 import taxi.model.Driver;
 
 public interface DriverService extends GenericService<Driver> {
+    Driver findByLogin(String login) throws AuthenticationException;
```
olekskov, line 7: Change the method signature to remove 'throws AuthenticationException' as it is not necessary for the 'findByLogin' method in the interface

src/main/java/taxi/service/DriverServiceImpl.java
```diff
@@ -12,6 +13,14 @@ public class DriverServiceImpl implements DriverService {
     @Inject
     private DriverDao driverDao;
 
+    @Override
+    public Driver findByLogin(String login) throws AuthenticationException {
+        return driverDao.findByLogin(login).orElseThrow(() ->
+                new AuthenticationException("The user with login "
+                        + login + " doesn't exist!")
+        );
+    }
```
olekskov, line 22: Please use exception class that extends RuntimeException and not tightly connected to authentication.
There may be a lot of scenarios when we try to find user that are not connected to authentication process at all
```suggestion
    public Driver findByLogin(String login) {
        return driverDao.findByLogin(login).orElseThrow(() ->
                new NoSuchElementException("The user with login "
                        + login + " doesn't exist!")
        );
    }
```

olekskov (changes requested):

olekskov (approved):
Good job

PR #627, opened 2023-06-26
Created an authentication procedure for driver

viprko:


src/main/java/taxi/dao/impl/DriverDaoImpl.java
```diff
@@ -37,7 +40,7 @@ public Driver create(Driver driver) {
 
     @Override
     public Optional<Driver> get(Long id) {
-        String query = "SELECT * FROM drivers WHERE id = ? AND is_deleted = FALSE";
+        String query = "SELECT * FROM drivers WHERE drivers.id = ? AND is_deleted = FALSE";
```
ykhmaruk, line 43: I think it is not a problem, but we have no join with any other table, so we don't need to specify driver.id, just id. You have the same point in other queries/

src/main/java/taxi/service/impl/AuthenticationServiceImpl.java
```diff
@@ -0,0 +1,23 @@
...
+import taxi.model.Driver;
+import taxi.service.AuthenticationService;
+import taxi.service.DriverService;
+
+@Service
+public class AuthenticationServiceImpl implements AuthenticationService {
+    @Inject
+    private DriverService driverService;
+
+    @Override public Driver login(String username, String password) throws AuthenticationException {
```
ykhmaruk, line 16: the annotation should be on the previous line

ykhmaruk (approved):
Good job! Just a few recommendations.

src/main/java/taxi/dao/impl/DriverDaoImpl.java
viprko: Yeah, after the "IDEA ultimate" install, an exception was thrown for SQL language, so I fixed it automatically and this happened.

src/main/java/taxi/controller/driver/LoginController.java
```diff
@@ -0,0 +1,42 @@
...
+            HttpSession session = req.getSession();
+            session.setAttribute("driver_id", user.getId());
+            session.setAttribute("authenticated", true);
+            resp.sendRedirect(req.getContextPath() + "/index");
+        } catch (AuthenticationException e) {
+            req.setAttribute("errorMsg", e.getMessage());
+            req.getRequestDispatcher("/WEB-INF/views/authentication/login.jsp").forward(req, resp);
+        }
+    }
+
```
maxline, line 41:
```suggestion
```

src/main/java/taxi/dao/impl/DriverDaoImpl.java
maxline: This is very strange, we should not have any exception here without using table name before `id` field. I will not mark this as mistake now, but I recommend to show this problem on QnA and try to fix.

src/main/java/taxi/service/impl/AuthenticationServiceImpl.java
maxline, line 16:
```suggestion
    @Override 
    public Driver login(String username, String password) throws AuthenticationException {
```

maxline (changes requested):
Very good job! I left a couple of minor comments. Please take a look

viprko:
Sorry for Checkstyle's issues. I experimenting with my IDEA settings for reformating code and sometimes missing a few mistakes

PR #638, opened 2023-06-30
add Solution

ArinaPolna:


src/main/java/taxi/controller/authentication/LoginController.java
```diff
@@ -0,0 +1,42 @@
...
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import javax.servlet.http.HttpSession;
+import taxi.exception.AuthenticationException;
+import taxi.lib.Injector;
+import taxi.model.Driver;
+import taxi.service.AuthenticationService;
+
+@WebServlet(urlPatterns = "/login")
```
AnastasiiaVariuchuk, line 15: it's cool that you decided to use @WebServlet, but then it's better to use it for other classes and clear web.xml from unnecessary

AnastasiiaVariuchuk (commented):
Good job, but let's improve it

AnastasiiaVariuchuk (approved):

sarakhmen (changes requested):
Good job but let's slightly improve your solution :)

maxline (approved):
Well done! Approved

PR #647, opened 2023-07-02
Jv web security hw

TymurProkhorov:


src/main/java/taxi/filter/AuthenticationFilter.java
```diff
@@ -0,0 +1,46 @@
...
+        Long driverId = (Long) session.getAttribute("driver_id");
+        if (driverId == null && allowedUrls.contains(req.getServletPath())) {
+            filterChain.doFilter(servletRequest, servletResponse);
+            return;
+        }
+        if (driverId == null) {
+            resp.sendRedirect("/login");
+            return;
+        }
+        filterChain.doFilter(servletRequest, servletResponse);
```
okuzan, line 44: [Both in AuthenticationServiceImpl and in AuthenticationFilter, try to use only one if block.
](https://mate-academy.github.io/jv-program-common-mistakes/java-web/web-security/java-web-security)

okuzan (changes requested):

src/main/java/taxi/filter/AuthenticationFilter.java
```diff
@@ -0,0 +1,42 @@
...
+            throws IOException, ServletException {
+        HttpServletRequest req = (HttpServletRequest) servletRequest;
+        HttpServletResponse resp = (HttpServletResponse) servletResponse;
+        HttpSession session = req.getSession();
+        Long driverId = (Long) session.getAttribute("driver_id");
+        if (!(driverId == null && allowedUrls.contains(req.getServletPath()))) {
+            resp.sendRedirect("/login");
+            return;
+        }
+        filterChain.doFilter(servletRequest, servletResponse);
```
olekskov, line 40: Please do not use ! on combined conditions. It's extremely hard to read and understand such structure.
If you remove ! you just need to switch code in if and else to get the same:
```suggestion
        if (driverId == null && allowedUrls.contains(req.getServletPath())) {
            filterChain.doFilter(servletRequest, servletResponse);
            return;
        }
        resp.sendRedirect(request.getContextPath() + "/login");
```

And now we see that this condition doesn't look like correct as we need following:
```suggestion
        if (driverId != null || allowedUrls.contains(req.getServletPath())) {
            filterChain.doFilter(servletRequest, servletResponse);
            return;
        }
        resp.sendRedirect(request.getContextPath() + "/login");
```
or
```suggestion
        if (driverId == null && !allowedUrls.contains(req.getServletPath())) {
            resp.sendRedirect(request.getContextPath() + "/login");
            return;
        }
        filterChain.doFilter(servletRequest, servletResponse);
```

olekskov (changes requested):

olekskov (approved):
Good job

PR #648, opened 2023-07-02
Hw web security

Fafakepa:


src/main/java/taxi/dao/CarDaoImpl.java
```diff
@@ -188,7 +188,7 @@ private void deleteAllDrivers(Car car) {
     }
 
     private List<Driver> getAllDriversByCarId(Long carId) {
```
volodymyrruso, line 190: you can move CarDaoImpl to package dao/impl but I'm not sure if it's necessary

src/main/java/taxi/service/DriverServiceImpl.java
```diff
@@ -38,4 +39,9 @@ public Driver update(Driver driver) {
     public boolean delete(Long id) {
         return driverDao.delete(id);
     }
+
```
volodymyrruso, line 42: same here with creating new package  service/impl and move impl class to impl package

volodymyrruso (approved):
Good job! left few minor comments, not sure if it's necessary so let's the mentor decide 👍

src/main/java/taxi/service/AuthenticationServiceImpl.java
```diff
@@ -0,0 +1,25 @@
...
+@Service
+public class AuthenticationServiceImpl implements AuthenticationService {
+    @Inject
+    private DriverService driverService;
+
+    @Override
+    public Driver login(String username, String password) throws AuthenticationException {
+        Optional<Driver> driver = driverService.findByLogin(username);
+        if (driver.isEmpty()) {
+            throw new AuthenticationException("Incorrect username or password, try again");
```
rovein, line 18:
```suggestion
            throw new AuthenticationException("Incorrect username, try again");
```
Here you can definitely say it, because the user was not found in DB by provided username.

rovein (approved):
Great job, see one comment.

PR #649, opened 2023-07-02
added solution

Fiasko3304:


m-pnsk (approved):
LGTM. But maybe it's better to put LoginController and LogOutController in a separate package?

rovein (approved):
Great job!

PR #650, opened 2023-07-02
controllers and filter created.

Hlib13:


src/main/java/taxi/service/AuthenticationServiceImpl.java
```diff
@@ -0,0 +1,28 @@
...
+
+    @Override
+    public Driver login(String login, String password) throws AuthenticationException {
+        Optional<Driver> driver = driverDao.findByLogin(login);
+        if (driver.isEmpty()) {
+            throw new AuthenticationException("Wrong login or password!");
+        }
+        if (driver.get().getPassword().equals(password)) {
+            return driver.get();
+        }
```
okuzan, line 25: do this with one if

okuzan (changes requested):

olekskov (approved):
Good job

PR #663, opened 2023-07-14
web security hw

RostyslavOnysh:


src/main/java/taxi/model/Driver.java
```diff
@@ -48,13 +79,25 @@ public boolean equals(Object o) {
...
+    }
+
+    @Override
+    public String toString() {
+        return "Driver{"
+                + "id=" + id
+                + ", name='" + name + '\''
+                + ", licenseNumber='" + licenseNumber + '\''
+                + ", login='" + login + '\''
+                + ", password='" + password + '\''
```
olekskov, line 100: Never include sensitive data like user's password in the toString() method. This can lead to unintended exposure of sensitive data.

src/main/java/taxi/controller/LogoutController.java
```diff
@@ -0,0 +1,18 @@
...
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+@WebServlet(urlPatterns = "/logout")
+public class LogoutController extends HttpServlet {
+    @Override
+    protected void doGet(HttpServletRequest req, HttpServletResponse resp)
+            throws ServletException, IOException {
+        req.getSession().invalidate();
+        req.getRequestDispatcher("/WEB-INF/views/login.jsp").forward(req,resp);
```
olekskov, line 16: After invalidating the session, you should redirect the user to the login page instead of forwarding. Forwarding keeps the original request URL in the browser address bar, which could lead to unexpected behavior if the user refreshes the page.

src/main/java/taxi/dao/CarDaoImpl.java
```diff
@@ -211,9 +211,13 @@ private Driver parseDriverFromResultSet(ResultSet resultSet) throws SQLException
         Long driverId = resultSet.getObject("id", Long.class);
         String name = resultSet.getString("name");
         String licenseNumber = resultSet.getString("license_number");
+        String login = resultSet.getString("login");
+        String password = resultSet.getString("password");
```
olekskov, line 215: You are trying to get the 'password' field from the result set, but 'password' field is not included in SQL query. You need to add 'password' in your SQL query.

src/main/java/taxi/service/AuthenticationServiceImpl.java
```diff
@@ -0,0 +1,23 @@
...
+
+@Service
+public class AuthenticationServiceImpl implements AuthenticationService {
+
+    @Inject
+    private DriverService driverService;
+
+    @Override
+    public Driver login(String login, String password) throws AuthenticationException {
+        Optional<Driver> driver = Optional.ofNullable(driverService.findByLogin(login));
```
olekskov, line 17: The 'Optional.ofNullable' call is unnecessary here because the 'driverService.findByLogin(login)' method should already return an Optional. Instead of using 'Optional.ofNullable', you should expect 'driverService.findByLogin(login)' to return an Optional.

src/main/java/taxi/controller/LoginController.java
```diff
@@ -0,0 +1,43 @@
...
+            throws ServletException, IOException {
+        String login = req.getParameter("login");
+        String password = req.getParameter("password");
+        try {
+            Driver driver = authenticationService.login(login,password);
+            HttpSession session = req.getSession();
+            session.setAttribute("driver_id", driver.getId());
+            resp.sendRedirect("/index");
+        } catch (AuthenticationException e) {
+            req.setAttribute("errorMsg",e.getMessage());
```
olekskov, line 39:
```suggestion
            req.setAttribute("errorMsg", e.getMessage());
```

src/main/java/taxi/controller/LoginController.java
```diff
@@ -0,0 +1,43 @@
...
+    @Override
+    protected void doPost(HttpServletRequest req, HttpServletResponse resp)
+            throws ServletException, IOException {
+        String login = req.getParameter("login");
+        String password = req.getParameter("password");
+        try {
+            Driver driver = authenticationService.login(login,password);
+            HttpSession session = req.getSession();
+            session.setAttribute("driver_id", driver.getId());
+            resp.sendRedirect("/index");
```
olekskov, line 37: Checklist item:
If you use sendRedirect() method in your controllers, please pass request.getContextPath() + "/your-endpoint" as a parameter. Currently, the context path is empty, but if it is not, your code still should work.

Please recheck all occurrences

src/main/java/taxi/model/Driver.java
```diff
@@ -48,13 +79,25 @@ public boolean equals(Object o) {
...
+        return Objects.equals(id, driver.id) && Objects.equals(name, driver.name)
+                && Objects.equals(licenseNumber, driver.licenseNumber)
+                && Objects.equals(login, driver.login)
+                && Objects.equals(password, driver.password);
     }
 
     @Override
     public int hashCode() {
-        return Objects.hash(id, name, licenseNumber);
+        return Objects.hash(id, name, licenseNumber, login, password);
```
olekskov, line 90: Do not include password into equals/hashCode methods. Further password will not be stored as plain text, it will be hashed and this may cause unpredictable behaviour

olekskov (changes requested):

Ivan95kos (approved):
Great job

PR #670, opened 2023-07-30
Authentication and loginService are realized

YaroslavKostenko89:


src/main/java/taxi/model/Driver.java
```diff
@@ -50,11 +75,13 @@ public boolean equals(Object o) {
         Driver driver = (Driver) o;
         return Objects.equals(id, driver.id)
                 && Objects.equals(name, driver.name)
-                && Objects.equals(licenseNumber, driver.licenseNumber);
+                && Objects.equals(licenseNumber, driver.licenseNumber)
+                && Objects.equals(login, driver.login)
+                && Objects.equals(password, driver.password);
```
olekskov, line 80: I would not recommend to add password to equals/hashcode methods.
Further it will be encrypted and depending on hashing algorithm this may cause improper comparison

olekskov (approved):
Good job.
Please pay attention to comment below for further tasks

PR #675, opened 2023-08-10
solution

VictoriaMrkvch:


src/main/java/taxi/service/AuthenticationServiceImpl.java
```diff
@@ -0,0 +1,28 @@
...
+        Optional<Driver> driver = driverDao.findByLogin(login);
+        try {
+            if (driver.get().getPassword().equals(password)) {
+                return driver.get();
+            }
+        } catch (Exception e) {
+            throw new DataProcessingException("Login or password was incorrect", e);
+        }
+        throw new DataProcessingException("Login or password was incorrect",
+                new RuntimeException());
```
aaananas, line 26: lets refactor this method to throw an exception in 1 place. You should not use DataProcessingException here. DataProcessingException is for dao layer

aaananas (changes requested):

okuzan (approved):

